PR #1, opened 2020-05-05
Add license

blankdots:
add default license, MIT


otahontas:
Looks good. I'm adding license info to readme and setup.py as well.

PR #10, opened 2020-05-12
Feature/make submissions queryable

otahontas:
### Description

This PR makes XML file submissions queryable from mongodb, changes module structure in api folder and adds some codestyle enhancement

These changes broke bunch of tests, so I'll add some commits later to this PR to bring up test coverage!

### Type of change

### Changes Made

- Codebase now has type hints (to be tested with mypy) and pep-style docstrings (tested with pydocstyle)
- Api is refactored to following modules: views (aiohttp endpoint methods), translator (takes submission actions such as adding or modifying and calls correct database operation) and parser (parses XML to JSON objects)
- All submission are parsed to JSON objects, JSON is submitted to submissions db and XML to backup db. Submission db can be searched with attributes from schemas, but backup db is searchable only with accession numbers and aliases.

### Testing


metadata_backend/api/parser.py
```diff
@@ -0,0 +1,239 @@
...
+    @staticmethod
+    def flatten_and_parse_json(data):
+        """Recursively flatten json and parse away unwanted characters.
+
+        :param data: JSON to be flattened
+        :returns: Flattened JSON
+        """
+        out = {}
+
+        def flatten(content, name=""):
```
blankdots, line 70: maybe a miss also here?
```suggestion
        def flatten(content: Union[Dict, List], name: str = "") -> None:
```
`Union` needs to be imported from typing

metadata_backend/api/parser.py
```diff
@@ -0,0 +1,239 @@
...
+
+        Currently just flattens JSON and generates random accession number as
+        placeholder (numbers are later on fetched from an another source).
+
+        :param xml_content: XML content as JSON
+        :returns: Parsed data as JSON
+        """
+        flattened_data = self.flatten_and_parse_json(data)
+        if "accession" not in flattened_data:
+            flattened_data["accession"] = random.randint(1, 10000)
```
blankdots, line 112: usually takes the form of `EGAD51775617678681` EGAD + some nb. for datasets or EGAF + some nb.
we can use  `'EGAD'+''.join(secrets.choice(string.digits) for i in range(16))` to be a a bit more close to what we will have.

`import secrets` is needed

see: https://docs.python.org/3/library/secrets.html

PR #12, opened 2020-05-19
Feature/add tox configs tests and fixes

### LINKED ISSUE #7: Add better support for testing asyncronous routines
#### Proposed solution
Some tests are currently skipped since async-await is not natively supported by unittest library. External testing library, such as `python-asyncio` could be used to fix this. See travis-report for more information: https://travis-ci.org/github/CSCfi/metadata_submitter/jobs/683816553

#### DoD (Definition of Done)
- Tests aren't skipped because of async-errors

otahontas:
### Description

(Rebased to state where previous PR was merged to develop).

This PR adds mypy and pydocstyle to tox and adds a couple of following smaller fixes:
- fixes some mypy-errors and adds missing type hints
- adds some try-catch checks for database and HTTP errors
- adds accession number generation
- makes it possible to submit multiple file same schema in one submission
- adds receipt generation
- adds some better submission.xml parsing

As discussed, receipt generator and submission.xml won't probably be used that much, and focus will be moved on to more rest-compliant endpoint features. Yet, they'll be now present in code base in case they're needed at some point (maybe for compatibility with EBI's service or what not). 

Tests are also now written in a way that doesn't result any async code running synchronously. Tests can be also now ran before each-commit with included pre-commit -hook. I'll add documentation for this soon.

### Related issues
Fixes #7 

### Type of change
- [x] This change requires a documentation update (following up in this PR)

### Changes Made

See git commit log below.

### Testing

- [x] Unit Tests


tox.ini
```diff
@@ -21,6 +21,22 @@ deps =
...
+    mypy
+# Mypy fails if 3rd party library doesn't have type hints configured.
+# Alternative to ignoring imports would be to write custom stub files, which
+# could be done at some point.
+commands = mypy --ignore-missing-imports metadata_backend/
+
+[testenv:pydocstyle]
+skip_install = true
+deps =
+    pydocstyle
```
blankdots, line 37: if you add https://pypi.org/project/flake8-docstrings/ as a dependency at the `testev:flake8` then this does not need to be done in a separate env.

tox.ini
```diff
@@ -1,13 +1,13 @@
 [tox]
-envlist = py{36,37},flake8,isort
+envlist = py{36,37},flake8,isort,mypy,pydocstyle
```
blankdots, line 2: should we add these to travis integration as well ?

tox.ini
blankdots: we will have a separate stage for generating and testing documentation, so this could be included with flake8

otahontas:
@blankdots I changed some code based on you comments. I also fixed bunch of typos from docstring (didn't notice them before), no effect on code itself. 

This should be now ok to merge.

PR #14, opened 2020-05-26
Feature/add more endpoints and queryable submissions

otahontas:
### Description

This PR adds some new enpoints and makes submission now more easily queryable from db. Querying enpoint is still under construction. 

There are also little tweaks here and there (try-except checks etc.).

### Type of change

### Changes Made

Folowing endpoints are added:
```
- /objects/
    - GET: Fetches all objects available (study, sample, analysis... etc)
- /objects/<schema>/
    - POST: Adds this type of object to database
- /objects/<schema>/<accessionId>
    - GET: Fetches a single object with its accessionId
        - If extended with ?format=xml, returns originally submitted xml file corresponding this accessionId
- /<schema>/<accessionId>
   - GET: Alias for /objects/<schema>/<accessionId>
```
### Testing

### DIFF
--- a/metadata_backend/api/parser.py
+++ b/metadata_backend/api/parser.py
@@ -1,11 +1,14 @@
 """Tool to parse XML files to JSON."""
 
+import re
 import secrets
 import string
-from typing import Dict, List, Union
+from datetime import datetime
+from typing import Dict, List
 
 from aiohttp import web
-from xmlschema import XMLSchema, XMLSchemaException
+from dateutil.relativedelta import relativedelta
+from xmlschema import AbderaConverter, XMLSchema, XMLSchemaException
 
 from ..helpers.schema_load import SchemaLoader, SchemaNotFoundException
 
@@ -13,7 +16,7 @@ from ..helpers.schema_load import SchemaLoader, SchemaNotFoundException
 class SubmissionXMLToJSONParser:
     """Methods to parse necessary data from different xml types.
 
-    Currently only submission-type is parsed excplicitly, others are parsed
+    Currently only submission-type is parsed explicitly, others are parsed
     just by flattening them.
     """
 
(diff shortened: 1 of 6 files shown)


metadata_backend/api/parser.py
```diff
@@ -234,21 +211,47 @@ def parse_submission(self, data: Dict) -> Dict:
...
-        schema = self.load_schema(xml_type)
-        self.validate(content, schema)
-        content_json_raw = schema.to_dict(content)[xml_type.upper()][0]
-        return getattr(self, f"parse_{xml_type}")(content_json_raw)
+        order = {"study": 1, "sample": 2, "experiment": 3, "run": 4,
+                 "analysis": 5, "dac": 6, "policy": 7, "dataset": 8,
+                 "project": 9}
+        return sorted(data, key=lambda x: order[x["schema"]])
+
+    def _to_lowercase(self, obj):
```
blankdots, line 240: missing typing

metadata_backend/api/translator.py
```diff
@@ -1,8 +1,8 @@
 """Handle translating API endpoint functionalities to CRUD operations."""
-
 from typing import Dict
 
 from aiohttp import web
+from bson import json_util
```
blankdots, line 5: more of a question: i see this is used just for `dumps` is there a difference between `json.dumps` ?

requirements.txt
```diff
@@ -1,5 +1,5 @@
 aiohttp
-gunicorn
 pymongo
+python-dateutil
```
blankdots, line 3: where is this used, should this be also in `setup.py` ?

why no more `gunicorn` ?

metadata_backend/api/translator.py
otahontas: Yeah, json.dumps crashes when trying to serialize data returned from Mongo. More info here: https://api.mongodb.com/python/current/api/bson/json_util.html

blankdots: ok. was just wondering.

requirements.txt
otahontas: Dateutil was added for [adding publish dates for studies](https://github.com/CSCfi/metadata-submitter/blob/origin/add_more_endpoints_and_queryable_submissions/metadata_backend/api/parser.py#L97) (I think ENA uses +2 months from submission date by default). I also think some date-related parsing is needed later to handle those HOLD/SUPPRESS etc actions, so this library is probably useful.

I recall we thought that gunicorn is not needed. But anyways, it should be removed from [deploy/app.sh](https://github.com/CSCfi/metadata-submitter/blob/origin/add_more_endpoints_and_queryable_submissions/deploy/app.sh#L7) as well, if it it's not used. What do you think?

otahontas: And setup.py gets parses requirements directly from requirements.txt: https://github.com/CSCfi/metadata-submitter/blob/master/setup.py#L5

blankdots: ok to both :)
In `deploy/app.sh` we need to see. I would still prefer to have it there for the moment, but don't remember exactly if the s2i container installs it automatically or not.

PR #28, opened 2020-05-29
Test/add test for parser.py

### LINKED ISSUE #19: Add better unit tests for parser.py module.
Try to get close to 90% coverage.

otahontas:
### Description

Adding tests for parser.py module and some fixes here and there.

I'm thinking that in the future we should create bunch of JSON test files, which are in format that we want them to be in db. Then parser.py can be just tested against these JSONs. 

### Issues
Fixes #19 

### Type of change

### Changes Made

- tests for parser.py-modules methods
- flake8-annotations, so missing type hints are now reported! (and obviously, some type hints we're missing, PR fixes those)
- wrote some new contribution tips, notably about tox and pre-commit possibilities. Also added misspell as configurable pre-commit, if someone wants to use it. 
- Parser.py had wrong error raising, fixed that.

### Testing
- [x] Unit Tests


scripts/pre-commit.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if [ $? -ne 0 ]; then
+    echo "=============================="
+    echo "Tests must pass before commit!"
+    echo "Note: Tox also checks non-staged changes, so you might need to stash
+    or add your changes before committing"
+    exit 1
+fi
+
+echo "Running misspell as a pre-commit hook"
```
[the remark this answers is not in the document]
otahontas: Fixed here: https://github.com/CSCfi/metadata-submitter/pull/28/commits/71eeca6cef039631edcfaecaae3ce85a89e8140a7

metadata_backend/api/parser.py
```diff
@@ -97,6 +95,7 @@ def _parse_sample(self, data: Dict) -> Dict:
         """Parse data from sample-type XML.
 
         Currently doesn't add anything extra.
+        ceva
```
blankdots, line 98:
```suggestion
```

blankdots: my bad

PR #31, opened 2020-06-02
Feature/improve database

otahontas:
### Description
Another small update, for database this time. 

### Type of change

### Changes Made
- Add missing crud operations (replace, update, delete)
- Add tests for crud operations
- Fix bug, where mongoclient was created individually each db_service object. MongoClient should be shared accross the whole application, since client handles pooling: [see more here](https://pymongo.readthedocs.io/en/stable/faq.html#how-does-connection-pooling-work-in-pymongo)

### Testing
- [x] Unit Tests

### DIFF
--- a/metadata_backend/database/db_services.py
+++ b/metadata_backend/database/db_services.py
@@ -2,8 +2,8 @@
 
 from typing import Dict
 
-from pymongo import errors
 from pymongo.cursor import Cursor
+from pymongo.errors import ConnectionFailure, OperationFailure, PyMongoError
 
 from ..conf.conf import db_client
 
@@ -42,7 +42,7 @@ class CRUDService:
         """
         try:
             db_service.database[collection].insert_one(document)
-        except errors.PyMongoError:
+        except (ConnectionFailure, OperationFailure, PyMongoError):
             raise
 
     @staticmethod
@@ -58,7 +58,7 @@ class CRUDService:
         """
         try:
             return db_service.database[collection].find(query)
-        except errors.ConnectionFailure:
+        except (ConnectionFailure, OperationFailure, PyMongoError):
             raise
 
     @staticmethod
@@ -78,7 +78,7 @@ class CRUDService:
             update_operation = {"$set": data_to_be_updated}
             db_service.database[collection].update_one(find_by_id_query,
                                                        update_operation)
-        except errors.ConnectionFailure:
+        except (ConnectionFailure, OperationFailure, PyMongoError):
             raise
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


metadata_backend/database/db_services.py
```diff
@@ -1,37 +1,35 @@
...
+found from deploy/mongodb.
+"""
 
 import os
 from typing import Dict
 
 from pymongo import MongoClient, errors
 from pymongo.cursor import Cursor
 
+# Set up database client
```
blankdots, line 17: how about we have a config module, to properly manage config ?

metadata_backend/database/db_services.py
```diff
@@ -77,3 +74,58 @@ def read(db_service: DBService, collection: str,
...
+        :param accession_id: Accession id for object to be updated
+        :param data_to_be_updated: JSON representing the data that replaces
+        old data
+        :raises: Error when read fails for any Mongodb related reason
+        """
+        try:
+            find_by_id_query = {"accessionId": accession_id}
+            db_service.database[collection].replace_one(find_by_id_query,
+                                                        data_to_be_updated)
+        except errors.ConnectionFailure:
```
blankdots, line 114: I think we should catch more that `ConnectionFailure`, maybe catch the specific ones first and then catch the general exception.

should be the same in `delete`.

also on `ConnectionFailure`, one should do a retry mechanism to retry the previous operation e.g. 10 times in 2-5 min after which raise an error (probably could be a later task -- add an issue if so).

metadata_backend/database/db_services.py
```diff
@@ -1,37 +1,35 @@
...
-        mongo_user = os.getenv("MONGO_INITDB_ROOT_PASSWORD", "admin")
-        mongo_password = os.getenv("MONGO_INITDB_ROOT_PASSWORD", "admin")
-        mongo_host = os.getenv("MONGODB_HOST", "localhost:27017")
-        url = f"mongodb://{mongo_user}:{mongo_password}@{mongo_host}"
-        self.client = MongoClient(url)
 
-
-class DBService(MongoClientCreator):
+class DBService:
     """Create database service used to communicate with database.
```
blankdots, line 26:
```suggestion
    """Create service used for database communication.
```

metadata_backend/database/db_services.py
otahontas: yeah, sounds reasonable. I'll add it

metadata_backend/database/db_services.py
otahontas: yeah, maybe we could catch ConnectionFailure, OperationFailure and then other pymongo errors.

Retry mechanism sounds like a task for later, I'll add issue for that.

PR #46, opened 2020-06-17
Metadata Submitter backend API - OpenAPI specs

### LINKED ISSUE #13: Add OpenAPI specs to project
#### Proposed solution
We should add OpenAPI specs when we have more api endpoints available. This can be added to project docs later to help with programmatic api usage. 

See more about spec here: https://swagger.io/specification/

#### DoD (Definition of Done)
- YAML file in repo specifying endpoints.

blankdots:
Adresses #13

### Description

View at: http://editor.swagger.io/?url=https://raw.githubusercontent.com/CSCfi/metadata-submitter/c76d7d711ed5cfdb59b494e740193c6110bfe25a/docs/specification.yml

For error messages we can also use: https://tools.ietf.org/html/rfc7807

For pagination we can use: https://tools.ietf.org/html/rfc5988 in the header the `next`, see https://developer.github.com/v3/#pagination

### Related issues

Fixes #13 

### Type of change

- [x] Documentation

### Testing

- [x] Tests do not apply

### Mentions

@csc-jm @otahontas what do you think about it ?

We can also extend the `schemas` endpoint where we will add new schemas or update them when we will support custom schemas.

### DIFF
--- a/docs/specification.yml
+++ b/docs/specification.yml
@@ -560,7 +560,7 @@ components:
         publishedDate:
           type: string
           description: datetime in ISO8601 format
-        accesstype:
+        accessType:
           type: string
           enum: ["PUBLIC", "REGISTERED", "CONTROLLED"]
           description: in order to determine if a dataset requires a user to be registered or it can be public or it requires special permission or controlled access
--- a/docs/specification.yml
+++ b/docs/specification.yml
@@ -279,6 +279,9 @@ paths:
             schema:
               type: string
               format: binary
+          application/json:
+              schema:
+                type: object
       responses:
         201:
           description: Created
@@ -345,6 +348,11 @@ paths:
       tags:
         - Manage
       summary: Update information on an object, on success return the updated object
+      requestBody:
+        content:
+          application/json:
+              schema:
+                type: object
       parameters:
       - name: schema
         in: path
@@ -387,6 +395,15 @@ paths:
       tags:
         - Manage
       summary: Update the whole object object, on success return the updated object
+      requestBody:
+        content:
+          multipart/form-data::
+            schema:
+              type: string
+              format: binary
+          application/json:
+              schema:
+                type: object
       parameters:
       - name: schema
         in: path
--- a/docs/specification.yml
+++ b/docs/specification.yml
@@ -98,7 +98,18 @@ paths:
     get:
       tags:
         - Query
-      summary: List of objects.
+      summary: List of objects. By default it will show 10 results per page.
+      parameters:
+      - in: query
+        name: page
+        schema:
+          type: string
+        description: Page number
+      - in: query
+        name: per_page
+        schema:
+          type: string
+        description: Results per page
       responses:
         200:
           description: OK
@@ -110,7 +121,7 @@ paths:
     get:
       tags:
         - Query
-      summary: Filter objects from a specific schema using various query string parameters
+      summary: Filter objects from a specific schema using various query string parameters.By default it will show 10 results per page.
       parameters:
       - name: schema
         in: path
@@ -118,6 +129,16 @@ paths:
         schema:
           type: string
         required: true
+      - in: query
+        name: page
+        schema:
+          type: string
+        description: Page number
+      - in: query
+        name: per_page
+        schema:
+          type: string
+        description: Results per page
       - in: query
         name: title
         schema:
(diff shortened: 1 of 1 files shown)


csc-jm:
/validate has nothing in the request body? 
/submit and /validate respond the same exact way?

Other than that all looks good to me

docs/specification.yml
```diff
@@ -0,0 +1,616 @@
...
+    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
+tags:
+  - name: Submission
+    description: Endpoints used in submission
+  - name: Query
+    description: Endpoints used for data retrieval
+  - name: Manage
+    description: Endpoints used for managing data
+paths:
+  /submit:
```
otahontas, line 19: Current version of `/submit` -endpoint allows insertion of multiple objects types at a time (e.g. study+sample+experiment), so response with single json accessionId doesn't reflect to what happened in db. Probably answer with json containing accessionId for each added object?

docs/specification.yml
```diff
@@ -0,0 +1,616 @@
...
+              schema:
+                $ref: "#/components/schemas/400BadRequest"
+        401:
+          description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/401Unauthorized"
+  /validate:
+    post:
```
otahontas, line 50: should have 
`
requestBody:
        content:
          multipart/form-data::
            schema:
              type: string
              format: binary
`

docs/specification.yml
```diff
@@ -0,0 +1,616 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/401Unauthorized"
+        403:
+          description: Forbidden
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/403Forbidden"
```
otahontas, line 246: 404 if nothing was found?

docs/specification.yml
```diff
@@ -0,0 +1,616 @@
...
+      parameters:
+      - name: schema
+        in: path
+        description: Title of the XML schema.
+        schema:
+          type: string
+        required: true
+      requestBody:
+        content:
+          multipart/form-data::
```
otahontas, line 260: can actually be json as well

docs/specification.yml
```diff
@@ -0,0 +1,616 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/401Unauthorized"
+        403:
+          description: Forbidden
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/403Forbidden"
```
otahontas, line 319: 404 if not found?

docs/specification.yml
```diff
@@ -0,0 +1,616 @@
...
+        schema:
+          type: string
+        required: true
+      - name: accessionId
+        in: path
+        description: filter objects in schema using accession ID
+        schema:
+          type: string
+        required: true
+      responses:
```
otahontas, line 337: patch should include json containing data that replaces old data?

docs/specification.yml
```diff
@@ -0,0 +1,616 @@
...
+        in: path
+        description: Unique id of the targeted service.
+        schema:
+          type: string
+        required: true
+      - name: accessionId
+        in: path
+        description: filter objects in schema using accession ID
+        schema:
+          type: string
```
otahontas, line 377: multipart/form-data and json for content-type?

docs/specification.yml
```diff
@@ -0,0 +1,616 @@
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/401Unauthorized"
+  /validate:
+    post:
+      tags:
+        - Submission
+      summary: Validate XML before submission.
+      responses:
+        201:
```
otahontas, line 55: Should be 200. Current response is `{"isValid: "boolean"}`

blankdots: added that + if that is `false` there should be a `detail` key (optional) with what went wrong.

docs/specification.yml
blankdots: yes, it can contain just the parts that needs replacing

docs/specification.yml
blankdots: good catch. changed to `ObjectsCreated` so that it also returns the schemas of the objects that were added.

blankdots:
@csc-jm and @otahontas good points addressed them in new commits, let me know of other changes.

docs/specification.yml
otahontas: yeah, but shouldn't we add it to api specs?

PR #47, opened 2020-06-17
Feature/add middlewares

### LINKED ISSUE #9: Add middlewares to handle http error raising and custom http errors
#### Proposed solution
Errors are currently handled and raised in [views.py,](https://github.com/CSCfi/metadata-submitter/blob/master/metadata_backend/api/views.py) but using middlewares would make error handling more modular. 

#### DoD (Definition of Done)
- http error handling is moved to middleware

#### Testing
- Can be done with integration tests later as we don't have them set up.
- unit tests: check if certain input return http error code

csc-jm:
### Description

Pretty standard implementation of http error middleware. Might have to adjust it later as right now it also converts all other exceptions into Internal Server Error, which might not be useful for the front-end.

### Related issues

Fixes #9 

### Type of change

### Changes Made

- add middleware in a separate file and reference it in server initiation

### Testing

- [x] Unit Tests

### DIFF
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -29,8 +29,8 @@ def error_middleware() -> Callable:
             return _json_exception(error.status, error)
 
         # If something else goes wrong, convert to Internal Server Error
-        except Exception as ex:
-            return _json_exception(500, ex)
+        # except Exception as ex:
+            # return _json_exception(500, ex)
 
     return http_error_handler
 
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -21,16 +21,9 @@ def error_middleware() -> Callable:
         """
         try:
             response = await handler(req)
-            if response.status == 404:
-                return _json_exception(response.status,
-                                       Exception(response.message))
             return response
         except HTTPError as error:
             return _json_exception(error.status, error)
-
-        # If something else goes wrong, convert to Internal Server Error
-        # except Exception as ex:
-            # return _json_exception(500, ex)
 
     return http_error_handler
 
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -1,0 +1,50 @@
+"""Middleware methods for server."""
+import json
+from typing import Callable
+
+from aiohttp.web import HTTPError, Request, Response, middleware
+
+
+def error_middleware() -> Callable:
+    """Middleware for handling exceptions recieved from the API methods.
+
+    :returns: Handled response
+    """
+
+    @middleware
+    async def http_error_handler(req: Request, handler) -> Response:
+        """Handle HTTP errors.
+
+        :param req: A request instance
+        :param handler: A request handler
+        :returns: JSON response for the error
+        """
+        try:
+            response = await handler(req)
+            if response.status == 404:
+                return _json_exception(response.status,
+                                       Exception(response.message))
+            return response
+        except HTTPError as error:
+            return _json_exception(error.status, error)
+
+        # If something else goes wrong, convert to Internal Server Error
+        except Exception as ex:
+            return _json_exception(500, ex)
+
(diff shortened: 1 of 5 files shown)


metadata_backend/api/middlewares.py
```diff
@@ -0,0 +1,50 @@
...
+    async def http_error_handler(req: Request, handler: Callable) -> Response:
+        """Handle HTTP errors.
+
+        :param req: A request instance
+        :param handler: A request handler
+        :returns: JSON response for the error
+        """
+        try:
+            response = await handler(req)
+            if response.status == 404:
```
otahontas, line 24: I don't think this is needed separately here, since 404's are also catched with except-clause below.

You can just `return await handler(req)`

metadata_backend/api/middlewares.py
```diff
@@ -0,0 +1,50 @@
...
+                return _json_exception(response.status,
+                                       Exception(response.message))
+            return response
+        except HTTPError as error:
+            return _json_exception(error.status, error)
+
+        # If something else goes wrong, convert to Internal Server Error
+        except Exception as ex:
```
otahontas, line 32: Like we discussed yesterday, this might hide some bugs and lead to weird situations. If program raises exceptions instead of httpErrors, we should add try-except to where the exception was raised.

metadata_backend/api/middlewares.py
```diff
@@ -0,0 +1,50 @@
...
+
+def _json_exception(status: int, exception: Exception) -> Response:
+    """Convert an HTTP exception into JSON response.
+
+    :param status: Status code of the HTTP exception
+    :param exception: Exception content
+    :returns: Response in JSON format
+    """
+    body = json.dumps({
+        'error': exception.__class__.__name__,
```
otahontas, line 46: Maybe better to return error as key and reason as value? I don't think class name is important for user and correct status code is anyways always returned.

tests/test_middlewares.py
```diff
@@ -0,0 +1,51 @@
...
+        """Test that middleware converts random error into a JSON response."""
+        response = await self.client.post("/submit", data=None)
+        # POST requesting nothing results in an assertion error
+        # which is not an HTTP error. Thus error status should be 500.
+        self.assertEqual(response.status, 500)
+        self.assertEqual(response.content_type, "application/json")
+        self.assertIn("detail", await response.json())
+
+    @unittest_run_loop
+    async def test_error_middleware_does_not_affect_good_responses(self):
```
otahontas, line 42: Probably no need to test this, since other tests would broke if middleware would affect them?

metadata_backend/api/middlewares.py
otahontas: Or we could actually add forward to frontend, where react renders 404 page?

otahontas:
I added some suggestions for stuff we discussed yesterday:
- removed /{schema}, we don't need it since we have /object/{schema}
- added checks that /object/{schema} returns 404 if schema is wrong (like "stuuufyy" etc)
- fixed that AssertionError bug @csc-jm noticed, it now returns 400 for error.

@csc-jm feel free to fix these, if you'd like to handle things differently!

metadata_backend/api/middlewares.py
blankdots: we will change this in the OpenAPI specs. so fo now I think returning the HTTP status (as redundant it is) would suffice.

otahontas:
Btw, would be also good to log errors in middleware

blankdots:
yeah logging is lacking a bit at the moment, was testing yesterday the backend, but logs were quite scarce

csc-jm:
Added a test for the 404 case in the handlers just to see for myself how that works. Can be removed if not a necessary test. Added some logging aspect in the middleware without being completely sure, if it was the proper way. Well, it's merged now. :D

blankdots:
@csc-jm you can create a new PR and we can merge that as well.

PR #51, opened 2020-06-18
Feature/add front build

### LINKED ISSUE #43: Add frontend build as servable static file to backend
#### Proposed solution
React supports building exportable production build with static files, which in turn can be served from backend. 

What we need:
- Handler for serving the static files. Check that it works with react router, which handles url routing inside react app.
- Dockerfile build process: maybe git clone from frontends master branch and build front as build-stage and copy it then to backend? 

#### DoD (Definition of Done)
Frontend can be served directly from aiohttp backend.

#### Testing
- probably end-to-end tests with built docker image. We can use Cypress or Selenium for this.

otahontas:
### Description

This PR adds frontend build to Dockerfile and adds necessary static routes to server.py.

### Related issues
Fixes #43 

### Type of change

### Changes Made

- See commit history for more details and comments.

### Testing

- [x] Tests do not apply (tested manually that it works)


otahontas:
@blankdots If I recall correctly, it's hard to get s2i to work when we have both node and python in building process? Should we just remove s2i-related info from docs?

blankdots:
yeah, s2i building 2 separate things going to he tricky, so yeah we can remove

blankdots:
a rebase might be needed

PR #60, opened 2020-06-25
Refactor/better error structure

### LINKED ISSUE #48: Redefine better error structure for front-end
#### Proposed solution
Regarding errors messages being displayed in the front-end, the error responses produced by the back-end might need to be restructured.

csc-jm:
### Description

Updating the way all http errors are handled so that the problem detail response is in accordance with [RFC 7807](https://tools.ietf.org/html/rfc7807).

### Related issues
Fixes #48 

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue) / refactoring

### Changes Made

- error handling middleware records and produces more detailed error responses
- edited validation error messages slightly for better readability

### Testing

- [x] Unit Tests (edited previous unittests)


csc-jm:
If this error structure is ok, I can edit the OpenAPI specs and include in this PR as well

metadata_backend/api/handlers.py
```diff
@@ -164,14 +164,16 @@ async def validate(self, req: Request) -> Response:
             raise web.HTTPBadRequest(reason=reason)
 
         except ParseError as error:
-            reason = f"Faulty XML file was given. Details: {error}"
-            body = json.dumps({"isValid": False, "reason": reason})
+            detail = f"Faulty XML file was given.<br/>ERROR: {error}"
```
otahontas, line 167: adding html-tags to responses is probably not a good idea, maybe just add real newline?

blankdots, line 167: I would omit the `<br/>` from the response. it is not necessary to format JSON responses.

also in line `173` below we would not need to replace it.

metadata_backend/api/handlers.py
```diff
@@ -164,14 +164,16 @@ async def validate(self, req: Request) -> Response:
...
+            body = json.dumps({"isValid": False, "detail": detail})
             return web.Response(body=body,
                                 content_type="application/json")
 
         except XMLSchemaValidationError as error:
-            reason = f"XML file is not valid. Details: {error}"
-            body = json.dumps({"isValid": False, "reason": reason})
+            error = str(error).replace('\n', '<br/>')  # for the frontend
```
otahontas, line 173: Same here, let's keep the original newlines. We should rather parse some information from validation error, maybe Reason and Instance. Maybe add issue about this, since this might be bigger task?

metadata_backend/api/handlers.py
```diff
@@ -164,14 +164,16 @@ async def validate(self, req: Request) -> Response:
...
+            detail = ("XML file is not valid against schema.<br/>"
```
otahontas, line 174: Same thing about `<br>`

metadata_backend/api/middlewares.py
```diff
@@ -25,22 +27,31 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
...
     :param status: Status code of the HTTP exception
     :param exception: Exception content
-    :returns: Response in JSON format
+    :param url: Request URL that caused the exception
+    :returns: Response in problem detail JSON format
     """
     body = json.dumps({
-        'error': status,
-        'detail': str(exception)
+        'type': "about:blank",
```
otahontas, line 48: maybe add link to swagger api schema here? It describes api responses a bit better than plain about:plank

blankdots:
seems that `coveralls` integration is broken, looking to fix it.

metadata_backend/api/handlers.py
otahontas:
```
Reason: invalid value 'Something wrong', it must be one of ['Whole Genome Sequencing', 'Metagenomics', 'Transcriptome Analysis', 'Resequencing', 'Epigenetics', 'Synthetic Genomics', 'Forensic or Paleo-genomics', 'Gene Regulation Study', 'Cancer Genomics', 'Population Genomics', 'RNASeq', 'Exome Sequencing', 'Pooled Clone Sequencing', 'Transcriptome Sequencing', 'Other']

Schema:

  <xs:enumeration xmlns:xs="http://www.w3.org/2001/XMLSchema" value="Whole Genome Sequencing">
    <xs:annotation>
      <xs:documentation>
         Sequencing of a single organism. 
      </xs:documentation>
    </xs:annotation>
  </xs:enumeration>

Instance:

  <STUDY_TYPE existing_study_type="Something wrong" />
```
We could try to parse necessary info from this

blankdots:
coveralls fixed in #61, that could be merged onto this.

metadata_backend/api/handlers.py
csc-jm: Dug through the xmlschema library source code and the last commit now includes method of getting the specific outputs for Reason and Instance as shown in your example. It's commented out however, in case we wanna parse through the errors differently

metadata_backend/api/middlewares.py
csc-jm: The type key is also optional so it could be left out for now. And if it's not included, its value is assumed to be 'about:blank'. The swagger api schema presents all possible errors so I'm not sure if it's that useful to include when describing a specific error.

blankdots: I think `about:blank` is ok for now until we will get a proper error schema.

metadata_backend/api/handlers.py
```diff
@@ -164,14 +164,20 @@ async def validate(self, req: Request) -> Response:
...
         except XMLSchemaValidationError as error:
-            reason = f"XML file is not valid. Details: {error}"
-            body = json.dumps({"isValid": False, "reason": reason})
+            '''
+            # Parsing the validation error message for reason and instance
+            from xml.etree import ElementTree
+            reason = error.reason
+            instance = ElementTree.tostring(error.elem, encoding="unicode")
+            '''
+            detail = f"XML file is not valid against schema.\nERROR: {error}"
```
blankdots, line 179:
```suggestion
            detail = f"XML file is not valid against schema.\nERROR: {error.reason}"
```

csc-jm: Yea, found that as well. The element where it's located (instance = `ElementTree.tostring(error.elem, encoding="unicode")`) might also be useful info to include

metadata_backend/api/handlers.py
otahontas: Good digging! Could you try how does that error look like in frontend?

metadata_backend/api/handlers.py
blankdots: yeap. :+1:

README.md
```diff
@@ -1,6 +1,9 @@
 ## Metadata submission service for SDA
 
-[![Build Status](https://travis-ci.org/CSCfi/metadata-submitter.svg?branch=master)](https://travis-ci.org/CSCfi/metadata-submitter)
+![Python Unit Tests](https://github.com/CSCfi/metadata-submitter/workflows/Python%20Unit%20Tests/badge.svg)
+![Integration Tests](https://github.com/CSCfi/metadata-submitter/workflows/Integration%20Tests/badge.svg)
+![Documentation Checks](https://github.com/CSCfi/metadata-submitter/workflows/Documentation%20Checks/badge.svg)
+![Python style check](https://github.com/CSCfi/metadata-submitter/workflows/Python%20style%20check/badge.svg)
 [![Coverage Status](https://coveralls.io/repos/github/CSCfi/metadata-submitter/badge.svg?branch=master)](https://coveralls.io/github/CSCfi/metadata-submitter?branch=master)
```
blankdots, line 7:
```suggestion
[![Coverage Status](https://coveralls.io/repos/github/CSCfi/metadata-submitter/badge.svg?branch=HEAD)](https://coveralls.io/github/CSCfi/metadata-submitter?branch=HEAD)
```

csc-jm:
After editing the OpenAPI specs, I remembered that errors should be in arrays. Thus the latest commit has the middleware insert the problem detail dicts into an 'errors' array so the response is similar to how it's in the specs. However I'm a bit unsure now if that's a good way to go about it.

csc-jm:
> line `179` from `handlers.py` could be: `detail = f"XML file is not valid against schema.\nERROR: {error.reason}"`

After testing around with the current front-end, it seems that even the value from `error.reason` by itself is in some cases too messy to be displayed in the field specified for error messages when submitting files. @otahontas and I discussed about things that could be done on the front-end side of things and determined that it might be more useful, if the validation endpoint instead returned an object like this:
```
{
"isValid": false,
"detail": { "reason": "reason for error",
            "instance": "element(s) that caused error"}
}
```
If this is too much to include in this PR and it should be dealt with in another issue, then the very last commit can be reverted and this PR should be good to merge.

PR #61, opened 2020-06-26
switch to github actions

blankdots:
### Related issues

Travis is annoying and github actions seem to work better

### Type of change

### Testing

- [x] Tests do not apply


.github/workflows/docs.yml
```diff
@@ -0,0 +1,41 @@
...
+      matrix:
+        os: [ubuntu-latest]
+    
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - name: Spell check install
+      run: curl -L https://git.io/misspell | bash
+    - name: Spell check docs
+      run: bin/misspell -error docs/*
```
otahontas, line 18: Should we run misspell also against codebase?

blankdots: sometimes it gives false positives on code. and we might use names that are not in English or acronyms so most likely no.

otahontas: Otherwise looks really fine

blankdots: plus documentation is more important.

otahontas: > sometimes it gives false positives on code. and we might use names that are not in English or acronyms so most likely no.

yeah that's okay. Our pre-commit hook (which I use) runs misspell against all non git-ignored files, but it can be fixed at some point if needed.

PR #62, opened 2020-06-26
Feature/add better xml parsing

### LINKED ISSUE #58: add better xml-json parsing process
#### Proposed solution

Try to improve our xml-json parsing process. Following things should be taken into account:
- get rid off acstract, title, descriptions newline characters and <>-characters.
- attribute-keys. Should be possible to remove attribute-keys and elevate values behind keys one level higher.
- children-keys. Conversion process now creates children-key and list of children. Maybe try to elevate these one step higher also?
- check if its possible to flatten json when its appropriate

#### DoD (Definition of Done)
- parsed jsons are more logical and easier to handle

#### Testing
- unit tests in parser module

otahontas:
### Description

This PR improves xml-json parsing in parser module, resulting cleaner jsons. Tried to focus also on some corner cases, but we need to test this with different xml files (other than just current test files). 

Differences between original conversion and couple of iterations I did: https://gist.github.com/otahontas/bc208d973ad8a2c707f071825ce2a967

Python script to create jsons from xml with server, used this for testing: https://gist.github.com/otahontas/bc208d973ad8a2c707f071825ce2a967#file-submit_and_get_jsons-py

### Related issues
Fixes #58 

### Type of change

### Changes Made

- Created new XMLConverter based on xmlschema-librarys AbderaConverter, also integrated CamelCase-parsing function
- Fixed query map to reflect new db structure
- Added docstrings to open whats happening in decoding process

### Testing

- [x] Needs more testing (start an issue or do a follow up PR about it) --> we need to write integration tests for query endpoints


metadata_backend/helpers/parser.py
```diff
@@ -1,41 +1,136 @@
...
+                    tmp_dict["value"] = children
+            return self.dict(tmp_dict)
+        else:
+            return children
+
+
 class XMLToJSONParser:
     """Methods to parse necessary data from different xml types."""
 
     def parse(self, type: str, content: str) -> Dict:
```
blankdots, line 123: `type` here needs some changing too to `schema_type`

otahontas:
> with this I think we also got our first candidate for pinning down a requirement: `xmlschema>=1.2.2`

I'll add this in some future PR

Issue #24, opened 2020-05-28
Add better unit tests for operator.py module.


otahontas:
Try to get close to 90% coverage.

PR #64, opened 2020-06-29
Feature/add delete and better integration tests

### LINKED ISSUE #55: Add possiblity to delete object recourses
#### Proposed solution

Add basic delete operation for deleting object from db.

#### DoD (Definition of Done)
deletion is possible

#### Testing
- is already unit tested at db level. Integration tests can be used for testing delete http req.

otahontas:
### Description

This PR makes it possible to delete objects from db, improves integration tests and adds more unit tests to operators module. Also some small fixes in handlers and operators modules.

### Related issues
Fixes #55 #24 

### Type of change

### Changes Made
- Added delete http method to /objects/{accessionId}/{schema} endpoint
- Operators catched all pymongoerrors with pokemon style, fixed it to catch only ConnectionFailure and OperationFailure. I'll improve pymongo-related error checking while changing pymongo to async version.
- Added more unittests for operators module, bumping up coverage to 86%
- Fixed integration test functions to work without affecting to each other. Previously this wasn't possible, because objects couldn't be deleted. 
- Add more logging to db-related operations
- Fixed multiple row strings to be presented with `"""`
- Added format-query to api specs
- Added check if schema type exists to query endpoint, fixed 404 tests to test only api-related urls, since frontend handles other 404 errors.
- Docker didn't build new versions from frontend master, fixed this.

### Testing
- [x] Unit Tests
- [x] Integration Tests


Dockerfile
```diff
@@ -2,9 +2,13 @@ FROM node:12-alpine as BUILD-FRONTEND
 
 RUN apk add --update \
     && apk add --no-cache git\
-    && rm -rf /var/cache/apk/* \
-    && git clone https://github.com/CSCfi/metadata-submitter-frontend
+    && rm -rf /var/cache/apk/*
+
+# Check if frotend in master has been updated. If there's no update, Docker
+# uses previous cache
+ADD https://api.github.com/repos/CSCfi/metadata-submitter-frontend/git/refs/heads/test-git-clone-docker-cache-workaround version.json
```
blankdots, line 9: not sure we should base this on github api ...

blankdots: maybe we should base it on tags ?

metadata_backend/api/handlers.py
```diff
@@ -234,15 +249,15 @@ async def _extract_xml_upload(req: Request, extract_one: bool = False
         if extract_one and files:
             reason = "Only one file can be sent to this endpoint at a time."
             raise web.HTTPBadRequest(reason=reason)
-        xml_type = part.name.lower()
-        if xml_type not in object_types:
-            raise web.HTTPBadRequest(reason="Not ok type")
+        type = part.name.lower()
```
blankdots, line 252: i think `type` is a reserved name in python so was maybe:
```suggestion
        schema_type = part.name.lower()
```

metadata_backend/api/handlers.py
```diff
@@ -234,15 +249,15 @@ async def _extract_xml_upload(req: Request, extract_one: bool = False
...
+        if type not in object_types:
```
blankdots, line 253:
```suggestion
        if schema_type not in object_types:
```

metadata_backend/api/handlers.py
```diff
@@ -234,15 +249,15 @@ async def _extract_xml_upload(req: Request, extract_one: bool = False
...
+            raise web.HTTPNotFound(reason=f"Theres no schema with name {type}")
```
blankdots, line 254:
```suggestion
            raise web.HTTPNotFound(reason=f"Theres no schema with name {schema_type}")
```

metadata_backend/api/handlers.py
```diff
@@ -234,15 +249,15 @@ async def _extract_xml_upload(req: Request, extract_one: bool = False
...
         data = []
         while True:
             chunk = await part.read_chunk()
             if not chunk:
                 break
             data.append(chunk)
         xml_content = ''.join(x.decode('UTF-8') for x in data)
-        files.append((xml_content, xml_type))
+        files.append((xml_content, type))
```
blankdots, line 262:
```suggestion
        files.append((xml_content, schema_type))
```

metadata_backend/api/operators.py
```diff
@@ -57,10 +58,34 @@ def create_metadata_object(self, type: str,
         """
         accession_id = self._generate_accession_id()
         self._handle_data_and_add_to_db(type, data, accession_id)
-        LOG.info(f"Inserting file to database succeeded: {type}, "
-                 f"{self.content_type}")
+        LOG.info(f"""Inserting file to database succeeded: {type}
```
blankdots, line 61: the same here did not notice the `type` - maybe `schema_type`

otahontas: yeah that's a fair point. I think type is also in other modules, so I'll go through the whole project to fix those.

Dockerfile
otahontas: true, let's use tags

PR #70, opened 2020-07-06
Feature/change from pymongo to motor

### LINKED ISSUE #32: Add retry mechanism for db failures and possibly change pymong to async version
#### Proposed solution
We should add retry mechanism for db operations in case of failures. Maybe if MongoDb ConnectionFailure happens, app could try 10 times in 2-5 min after which it will raise an error. Check Mongodb docs for retryable writes: https://docs.mongodb.com/manual/core/retryable-writes/

When implementig this, remember to check for aiohttp shielding, since we don't want aiohttp to cancel awaiting coroutine while retry mechanism is doing its task. See more: https://docs.aiohttp.org/en/stable/web_advanced.html?highlight=shield#web-handler-cancellation

#### DoD (Definition of Done)
- There is reliable connection failure mechanism when doing db writes.

#### Testing
- Maybe stress test this one? Not sure, need to research.

otahontas:
### Description

This PR changes underlying database operations to be coroutine-based (by replacing Pymongo with Motor) and adds retryable database operations. 

### Related issues
Fixes #32 

### Type of change
- [x] Refactor

### Changes Made
- Change operations in db_services module to be async
- Change operations in operators module to be async
- Change calls for operators in handlers module to be async
- Fix database client sharing to work through aiohttp app, recommended by aiohttp docs, see: https://docs.aiohttp.org/en/stable/web_advanced.html#data-sharing-aka-no-singletons-please (otherwise motor is not always using the same loop than aiohttp which creates problems)
- Refactor away some shared code in methods in operators module
- Fix tests for db, operators and handlers, remove mongomock library since it supports only blocking operations. 

### Testing

- [x] Unit Tests

### DIFF
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -9,8 +9,8 @@ from typing import Any, Dict, List, Tuple, Union
 from aiohttp import web
 from bson import json_util
 from dateutil.relativedelta import relativedelta
+from motor.motor_asyncio import AsyncIOMotorClient, AsyncIOMotorCursor
 from multidict import MultiDictProxy
-from motor.motor_asyncio import AsyncIOMotorCursor, AsyncIOMotorClient
 from pymongo.errors import ConnectionFailure, OperationFailure
 
 from ..conf.conf import query_map
@@ -153,7 +153,7 @@ class Operator(BaseOperator):
     Operations are implemented with JSON format.
     """
 
-    def __init__(self, db_client) -> None:
+    def __init__(self, db_client: AsyncIOMotorClient) -> None:
         """Initialize database and content-type.
 
         :param db_client: Motor client used for database connections. Should be
@@ -277,7 +277,7 @@ class XMLOperator(BaseOperator):
     Operations are implemented with XML format.
     """
 
-    def __init__(self, db_client) -> None:
+    def __init__(self, db_client: AsyncIOMotorClient) -> None:
         """Initialize database and content-type.
 
         :param db_client: Motor client used for database connections. Should be
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -52,8 +52,8 @@ class BaseOperator(ABC):
         accession_id = (await self.
                         _format_data_to_create_and_add_to_db(schema_type,
                                                              data))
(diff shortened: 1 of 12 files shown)


metadata_backend/api/operators.py
```diff
@@ -25,87 +25,123 @@ class BaseOperator(ABC):
...
+
+        :param schema_type: Schema type of the object to read.
+        :param data: Data to be saved to database.
+        :returns: Accession id for the object inserted to database
+        """
+        accession_id = (await self.
+                        _format_data_to_create_and_add_to_db(schema_type,
+                                                             data))
+        LOG.info(f"Inserting file with schema {schema_type} to database "
+                 f"succeeded with accession id: {accession_id}")
```
blankdots, line 56: maybe this can be multiline with `"""` ?

otahontas: This is done to avoid adding linebreak, which multiline does.

metadata_backend/database/db_service.py
```diff
@@ -1,11 +1,46 @@
...
+        max_attempts = 5
+        for attempt in range(1, max_attempts + 1):
+            try:
+                return await db_func(*args, **kwargs)
+            except AutoReconnect:
+                if attempt == 5:
+                    message = (f"Connection to database failed after {attempt}"
+                               "tries")
+                    raise ConnectionFailure(message=message)
+                LOG.info("Connection not successful, trying to reconnect."
```
blankdots, line 37: multi line `"""` and maybe this is `LOG.error` ?

metadata_backend/api/operators.py
```diff
@@ -25,87 +25,123 @@ class BaseOperator(ABC):
...
+        Data formatting to JSON or XML must be implemented by corresponding
+        subclass.
+
         :param schema_type: Schema type of the object to read.
         :param accession_id: Accession Id of the object to read.
         :raises: 400 if reading was not succesful, 404 if no data found
         """
         try:
-            data_raw = self.db_service.read(schema_type, accession_id)
+            data_raw = await self.db_service.read(schema_type, accession_id)
```
blankdots, line 71: should we add some logs here ?

metadata_backend/api/operators.py
blankdots: ok

PR #74, opened 2020-07-07
Feature/add pagination

### LINKED ISSUE #54: Add pagination to api endpoints
#### Proposed solution
Allow pagination of queries based on some preset page size (5,10,20 maybe?). See more: https://www.codementor.io/@arpitbhayani/fast-and-efficient-pagination-in-mongodb-9095flbqr#approach-1-using-cursorskip-and-cursorlimit

#### DoD (Definition of Done)
Responses are paginated

#### Testing
Unit / integration

otahontas:
### Description

This PR adds pagination to query responses, following our api specs.

### Related issues
Fixes #54 

### Type of change

### Changes Made
- Added pagination parameter checks and pagination response to handlers module
- Added pagination logic to operators module
- Added db operations for size of query results (yepp, this can't be done directly with cursor...) 
- Moved away json dumps from operators module --> now operators are only returning strings, lists or dicts, which can then be dumped to json / xml in handlers module. Makes program logic clearer imo.

+ added some tiny fixes I noticed to handlers errors 

### Testing
- [x] Unit Tests
- [x] Integration Tests

### Mentions
Are we ok with this or do we wan't to include for example link headers (see https://stackoverflow.com/questions/12168624/pagination-response-payload-from-a-restful-api)?

### DIFF
--- a/tests/test_db_service.py
+++ b/tests/test_db_service.py
@@ -4,6 +4,7 @@ from unittest.mock import MagicMock, patch
 from pymongo.results import InsertOneResult, UpdateResult, DeleteResult
 from pymongo.errors import AutoReconnect, ConnectionFailure
 from bson import ObjectId
+from motor.motor_asyncio import AsyncIOMotorCursor
 
 from metadata_backend.database.db_service import DBService
 
@@ -100,6 +101,13 @@ class DatabaseTestCase(AsyncTestCase):
         self.collection.delete_one.assert_called_once_with({"accessionId":
                                                            self.id_stub})
         assert success
+
+    def test_query_executes_find(self):
+        """Test that find is executed, so cursor is returned."""
+        self.collection.find.return_value = AsyncIOMotorCursor(None, None)
+        cursor = self.test_service.query("test", {})
+        assert type(cursor) == AsyncIOMotorCursor
+        self.collection.find.assert_called_once_with({})
 
     async def test_db_operation_is_retried_with_increasing_interval(self):
         """Patch timeout to be 0 sec instead of default, test autoreconnect."""
--- a/tests/test_db_service.py
+++ b/tests/test_db_service.py
@@ -39,7 +39,7 @@ class DatabaseTestCase(AsyncTestCase):
         """Test client is shared across different db_service_objects."""
         foo = DBService("foo", self.client)
         bar = DBService("bar", self.client)
-        self.assertEqual(foo.db_client, bar.db_client)
+        self.assertIs(foo.db_client, bar.db_client)
 
     async def test_create_inserts_data(self):
         """Test that create method works and returns success."""
--- a/tests/test_db_service.py
+++ b/tests/test_db_service.py
@@ -97,7 +97,7 @@ class DatabaseTestCase(AsyncTestCase):
         self.collection.delete_one.return_value = futurized(
             DeleteResult({}, True)
         )
-        success = await self.test_service.delete("yolo", self.id_stub)
+        success = await self.test_service.delete("test", self.id_stub)
         self.collection.delete_one.assert_called_once_with({"accessionId":
                                                            self.id_stub})
         assert success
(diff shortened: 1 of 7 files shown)


tests/test_db_service.py
```diff
@@ -96,11 +97,25 @@ async def test_delete_deletes_data(self):
...
+        success = await self.test_service.delete("test", self.id_stub)
         self.collection.delete_one.assert_called_once_with({"accessionId":
                                                            self.id_stub})
         assert success
 
+    def test_query_executes_find(self):
+        """Test that find is executed, so cursor is returned."""
+        self.collection.find.return_value = AsyncIOMotorCursor(None, None)
+        cursor = self.test_service.query("test", {})
+        assert type(cursor) == AsyncIOMotorCursor
```
blankdots, line 109: maybe `self.assertEquals` ? using an `AsyncTestCase` also does not seem to be the only place. look at https://docs.python.org/3.7/library/unittest.html for a complete list of what could be used.

tests/test_handlers.py
```diff
@@ -205,7 +200,7 @@ async def test_get_object(self):
         response = await self.client.get(url)
         assert response.status == 200
```
blankdots, line 201: a bit weird to use `assert` in one place and `self.assertEquals` in the next one :confused:

tests/test_db_service.py
otahontas: I'm not following - how does `AsyncTestCase` relate to `assertEquals` here?

Also, why assertEquals instead of assert? I think assert makes code more readable.

tests/test_handlers.py
otahontas: Yeah that I agree, which way to go? Assert or assertEquals?

tests/test_db_service.py
blankdots: it is not about readable code it is about using the `unittest.Testcase` class methods which are inherited: https://github.com/kwarunek/aiounittest/blob/master/aiounittest/case.py#L6 
`assert` has a place but it is not in a unit test class

tests/test_handlers.py
blankdots: `assertEquals` as we are using the unit test class

tests/test_db_service.py
blankdots: https://stackoverflow.com/questions/48216473/using-asserttrue-vs-assertequal-in-unittest

otahontas: Fair point. But what do you mean with this: "using an AsyncTestCase also does not seem to be the only place"?

blankdots: missing punctuation my bad. I meant that this is not the only place you used `assert` instead of `assertEquals` or similar
comment updated.

otahontas: Ah okay. Ok if I make another PR to change `assert`s?

blankdots: well ok .. I will review the rest of the PR tomorrow.

otahontas: Wasn't actually that big of a deal, so I included assert fixes in this PR

otahontas:
> Looks great, or at least I didn't catch any faults. The link headers would seemingly make things easier down the line right? Might be worth implementing or at least issueing for later.

Added issue for links: #75

tests/test_db_service.py
```diff
@@ -38,7 +39,7 @@ def test_db_services_share_mongodb_client(self):
         """Test client is shared across different db_service_objects."""
         foo = DBService("foo", self.client)
         bar = DBService("bar", self.client)
-        assert foo.db_client is bar.db_client
+        self.assertEqual(foo.db_client, bar.db_client)
```
blankdots, line 42:
```suggestion
        self.assertIs(foo.db_client, bar.db_client)
```
moving `is` into `equal` does not seem like the right way here

PR #81, opened 2020-07-08
Address logging

### LINKED ISSUE #72: Review logging and logging levels
#### Description

if there is an error use `.error` log if there otherwise use `.info` or `.debug` for additional information

#### DoD (Definition of Done)

review logging for existing source code.

#### Testing

Peer review

blankdots:
### Related issues

Fixes #72 

### Type of change

### Changes Made

Added error, info and debug logs to functions

### Testing

- [x] Tests do not apply

### Mentions

debug log might be verbose but it is good for tracing if functions were done.

### DIFF
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -74,6 +74,7 @@ class BaseOperator(ABC):
             data = await self._format_read_data(schema_type, data_raw)
         except (ConnectionFailure, OperationFailure) as error:
             reason = f"Error happened while getting file: {error}"
+            LOG.error(reason)
             raise web.HTTPBadRequest(reason=reason)
         return data, self.content_type
 
@@ -108,11 +109,13 @@ class BaseOperator(ABC):
             insert_success = (await self.db_service.create(schema_type, data))
         except (ConnectionFailure, OperationFailure) as error:
             reason = f"Error happened while getting file: {error}"
+            LOG.error(reason)
             raise web.HTTPBadRequest(reason=reason)
         if insert_success:
             return data["accessionId"]
         else:
-            reason = "Inserting file to database failed for some reason."
+            reason = "Inserting file to database failed for some (un)known reason."
+            LOG.error(reason)
             raise web.HTTPBadRequest(reason=reason)
 
     async def _remove_object_from_db(self,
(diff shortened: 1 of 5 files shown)


metadata_backend/api/operators.py
```diff
@@ -230,6 +243,7 @@ async def _format_data_to_create_and_add_to_db(self, schema_type: str,
         data["dateModified"] = datetime.utcnow()
         if schema_type == "study":
             data["publishDate"] = datetime.utcnow() + relativedelta(months=2)
+        LOG.debug(f"Operator format data for {schema_type} to add to DB")
```
otahontas, line 246: format -> formatted

metadata_backend/api/operators.py
```diff
@@ -315,6 +330,7 @@ async def _format_data_to_create_and_add_to_db(self, schema_type: str,
         accession_id = (await Operator(db_client).
                         _format_data_to_create_and_add_to_db(schema_type,
                                                              data_as_json))
+        LOG.debug(f"XMLOperator format data for {schema_type} to add to DB")
```
otahontas, line 333: format -> formatted

metadata_backend/database/db_service.py
```diff
@@ -142,6 +147,7 @@ def query(self, collection: str, query: Dict) -> AsyncIOMotorCursor:
         :param query: query to be used
         :returns: Async cursor instance which should be awaited when iterating
         """
+        LOG.debug("DB doc query performed.")
```
otahontas, line 150: Query doesn't actually happen here, it is done when data is returned from function here: https://github.com/CSCfi/metadata-submitter/blob/ea72ccddf7c9cfbd1aa2833adf773a7e6e8ba8d8/metadata_backend/api/operators.py#L213

Maybe move logging there?

metadata_backend/api/operators.py
blankdots: done.

metadata_backend/api/operators.py
blankdots: done.

metadata_backend/database/db_service.py
blankdots: it was meant to signal that a query was called, but will add logging there as suggested.

blankdots: a bit confused on what to add there because there is already logs ... :confused:

otahontas: you're right, my bad. This is ok now

PR #93, opened 2020-07-16
Refactor: Extract validation error handling to different module

### LINKED ISSUE #66: Extract validation error handling to different module
#### Proposed solution
Currently xml-file validation is handled in three different endpoints: /submit, /objects/{schema} and /validate. It's probably good to refactor validation process and http error raising to different module, so users can expect same type of json-responses regardless of endpoint they're using. 

#### DoD (Definition of Done)
We don't have copypaste code regarding xml-file validation.

#### Testing
- unit tests

csc-jm:
### Description

PR adds a `validator.py` to helper tools and readjusts `/validate` endpoint and `parser.py` to call the new method instead. No new tests added yet because previous tests work fine. If this refactoring looks ok though, previous tests dealing with validation could be moved to a new test class.

### Related issues

Fixes #66 

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue) / refactoring

### Changes Made

- Adds helper tool for validating XMLs
- removes previous validation calls and redirects those calls to the new validation method
- edited 2 of the tests in `test_parser.py` that do not apply anymore
- edited some logging calls and fixed a typo I spotted

### Testing

- [x] Needs testing (?)

### DIFF
--- a/metadata_backend/helpers/validator.py
+++ b/metadata_backend/helpers/validator.py
@@ -1,0 +1,61 @@
+"""Utility class for validating XML files."""
+
+import re
+import json
+from io import StringIO
+from urllib.error import URLError
+
+from aiohttp import web
+from xmlschema import XMLSchema, XMLSchemaValidationError
+from xmlschema.etree import ElementTree, ParseError
+
+from ..helpers.logger import LOG
+
+
+class XMLValidator:
+    """Validator implementation."""
+
+    def __init__(self, schema: XMLSchema, xml: str) -> None:
+        """Variables."""
+        self.schema = schema
+        self.xml_content = xml
+        self.resp_body = self.get_validation()
+
+    def get_validation(self) -> str:
+        """Where validation happens."""
+        try:
+            self.schema.validate(self.xml_content)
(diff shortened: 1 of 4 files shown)


metadata_backend/helpers/validator.py
```diff
@@ -0,0 +1,85 @@
...
+    def __init__(self, schema: XMLSchema, xml: str) -> None:
+        """Set Variables and initiate validation.
+
+        :param schema: Schema to be used
+        :param content: Content of XML file to be validated
+        """
+        self.schema = schema
+        self.xml_content = xml
+        self.resp_body = self.get_validation()
+        self.xmlIsValid = self.is_valid()
```
blankdots, line 27: wondering if making use of `@property` would have been useful :thinking:

otahontas: yeah I'll second this, maybe rename `get_validation() -> resp_body()` and then decorate `resp_body()` and `is_valid()` with `@property`

otahontas:
very readable, good work!

PR #95, opened 2020-07-21
Add authentication/authorization middleware

### LINKED ISSUE #44: Add Authentication and Authorization middleware
#### Proposed solution

We need to be able to authenticate and authorize users that use the API.
We need to check if the users are authenticated before submitting (first iteration can have just authentication for submission).

Certain submission related information might not be published or released or provided under control access that is where the authorization part comes into play. We need to make the necessary changes in DB and the API to allow access to that information only if the user is authorised.

We will working with oAuth2 and JWT.
Optional support Elixir AAI out of the box.

Possible libraries: https://github.com/lepture/authlib

#### DoD (Definition of Done)

Middleware that checks the user credentials and returns appropriate error if unauthorised or forbidden.

#### Testing

Unit testing with the possibility to add integration testing later.

csc-jm:
### Description

PR includes a middleware for validating and authenticating a JWT included in the request headers, similar in structure to the one found in [Beacon API Server](https://github.com/CSCfi/beacon-python/blob/dev/beacon_api/utils/validate_jwt.py). Things will still need to be added on to this middleware, for example #96.

### Related issues

Fixes #44 

### Type of change

### Changes Made

- added authentication middleware implementation 
- adds unit tests and includes helper methods in `test_middleware.py`

### Testing

- [x] Unit Tests

### DIFF
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -75,7 +75,6 @@ def jwt_middleware() -> Callable:
             if not re.match('Bearer', scheme):
                 raise web.HTTPUnauthorized(reason="Invalid token scheme, "
                                                   "Bearer required.")
-
             if token is None:
                 raise web.HTTPUnauthorized(reason='Token cannot be empty.')
 
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -73,15 +73,15 @@ async def jwt_authentication(req: Request, handler: Callable) -> Response:
 
         # JWK and JWTClaims parameters for decoding
         key = environ.get('PUBLIC_KEY', None)
-        # TODO: get Oauth2 public key
-        # TODO: verify audience claim
+
+        # Get Oauth2 public key and verify audience claim here
 
         # Include claims that are required to be present
         # in the payload of the token
         claims_options = {
             "iss": {
                 "essential": True,
-                # TODO: add the proper issuer URLs here
+                # Proper issuer URLs need to be added as the values
                 "values": ["haka_iss", "elixir_iss"]
             },
             "exp": {
@@ -95,7 +95,7 @@ async def jwt_authentication(req: Request, handler: Callable) -> Response:
             claims.validate()
             LOG.info('Auth token decoded and validated.')
 
-            # TODO: Retrieve GA4GH passports and process into permissions
+            # Retrieve GA4GH passports and process into permissions here
 
             req["token"] = {"authenticated": True}
             return await handler(req)
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -67,7 +67,7 @@ def jwt_middleware() -> Callable:
             # Check token exists
             try:
                 scheme, token = req.headers.get('Authorization').split(' ')
-                LOG.info('Auth Token Received.')
+                LOG.info('Auth token received.')
             except Exception as err:
                 raise web.HTTPUnauthorized(reason=str(err))
 
@@ -96,7 +96,9 @@ def jwt_middleware() -> Callable:
                 claims = jwt.decode(token, key, claims_options=claims_options)
                 claims.validate()
                 LOG.info('Auth token decoded and validated.')
+
                 # TODO: Retrieve GA4GH passports and process into permissions
+
                 req["token"] = {"authenticated": True}
                 return await handler(req)
             except errors.MissingClaimError as err:
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -46,6 +46,7 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
 
     return http_error_handler
 
+
 @middleware
 async def jwt_authentication(req: Request, handler: Callable) -> Response:
     """Middleware for validating and authenticating JSON web token.
@@ -60,8 +61,8 @@ async def jwt_authentication(req: Request, handler: Callable) -> Response:
         try:
             scheme, token = req.headers.get('Authorization').split(' ')
             LOG.info('Auth token received.')
-        except Exception as err:
-            raise web.HTTPUnauthorized(reason=str(err))
+        except ValueError as err:
+            raise web.HTTPUnauthorized(reason=f"Failure to read token: {err}")
 
         # Check token has proper scheme and was provided.
         if not re.match('Bearer', scheme):
@@ -113,7 +114,6 @@ async def jwt_authentication(req: Request, handler: Callable) -> Response:
     else:
         req["token"] = {"authenticated": False}
         return await handler(req)
-
 
 
 def _json_exception(status: int, exception: web.HTTPException,
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -13,47 +13,41 @@ from yarl import URL
 from ..helpers.logger import LOG
 
 
-def error_middleware() -> Callable:
+@middleware
+async def http_error_handler(req: Request, handler: Callable) -> Response:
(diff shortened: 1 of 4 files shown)


metadata_backend/api/middlewares.py
```diff
@@ -50,6 +53,78 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
     return http_error_handler
 
 
+def jwt_middleware() -> Callable:
```
otahontas, line 56: We can just import `jwt_authentication` and add `middlewares=[jtw_authentication ]` in server.py, no need to wrap it inside another function. (This actually applies to error middleware as well).

metadata_backend/api/middlewares.py
```diff
@@ -50,6 +53,78 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
...
+    :returns: Successful requests unaffected
+    """
+    @middleware
+    async def jwt_authentication(req: Request, handler: Callable) -> Response:
+        if 'Authorization' in req.headers:
+            # Check token exists
+            try:
+                scheme, token = req.headers.get('Authorization').split(' ')
+                LOG.info('Auth token received.')
+            except Exception as err:
```
otahontas, line 71: catching every Exception doesn't seem wise, maybe specify what exceptions you're trying catch?

metadata_backend/api/middlewares.py
```diff
@@ -50,6 +53,78 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
...
+            # Check token has proper scheme and was provided.
+            if not re.match('Bearer', scheme):
+                raise web.HTTPUnauthorized(reason="Invalid token scheme, "
+                                                  "Bearer required.")
+            if token is None:
+                raise web.HTTPUnauthorized(reason='Token cannot be empty.')
+
+            # JWK and JWTClaims parameters for decoding
+            key = environ.get('PUBLIC_KEY', None)
+            # TODO: get Oauth2 public key
```
otahontas, line 83: Remove TODOs, add issues instead

metadata_backend/api/middlewares.py
```diff
@@ -50,6 +53,78 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
...
+            # TODO: verify audience claim
+
+            # Include claims that are required to be present
+            # in the payload of the token
+            claims_options = {
+                "iss": {
+                    "essential": True,
+                    # TODO: add the proper issuer URLs here
```
otahontas, line 91: same thing here

metadata_backend/api/middlewares.py
```diff
@@ -50,6 +53,78 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
...
+                raise web.HTTPUnauthorized(reason="Token signature is invalid"
+                                                  f", {err}")
+            except errors.InvalidTokenError as err:
+                raise web.HTTPUnauthorized(reason="Invalid authorization token"
+                                                  f": {err}")
+        else:
+            req["token"] = {"authenticated": False}
+            return await handler(req)
+
+    return jwt_authentication
```
otahontas, line 125: no need to return this, if you import jwt_authentication in servers.py (see previous comment)

tests/test_middlewares.py
```diff
@@ -13,15 +16,45 @@ async def get_application(self):
...
+            'name': "tester",
+            'iss': "haka_iss",
+            'exp': 9999999999
+        }
+        self.pem = {
+            'kty': "oct",
+            'alg': "RS256",
+            'k': "GawgguFyGrWKav7AX4VKUg"
+        }
+        os.environ['PUBLIC_KEY'] = json.dumps(self.pem)
```
otahontas, line 36: setUp is run for all middleware tests, so setting env variables here also affets non-authentication tests. I'd suggest that you isolate authentication tests to new testcase class.

tests/test_middlewares.py
```diff
@@ -13,15 +16,45 @@ async def get_application(self):
...
+
+    async def tearDownAsync(self):
+        """Unset the public key."""
+        os.unsetenv('PUBLIC_KEY')
```
otahontas, line 40: Same thing for tearDown here

metadata_backend/api/middlewares.py
```diff
@@ -50,6 +53,78 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
...
+                }
+            }
+
+            # Decode and validate token
+            try:
+                claims = jwt.decode(token, key, claims_options=claims_options)
+                claims.validate()
+                LOG.info('Auth token decoded and validated.')
+
+                # TODO: Retrieve GA4GH passports and process into permissions
```
blankdots, line 105: this is done here: https://github.com/CSCfi/beacon-python/blob/dev/beacon_api/permissions/ga4gh.py

csc-jm: Added the issue #96, which also mentions this.

metadata_backend/api/middlewares.py
csc-jm: Upon checking the aiohttp documentation, I now recognize this is a good point. I guess the original idea with this  was that there might be multiple methods under the `error_middleware()` so that `server = web.Application(middlewares=[error_middleware()]` in `server.py`, would access them all. That never ended being the case, so good catch!

metadata_backend/api/middlewares.py
```diff
@@ -73,15 +73,15 @@ async def jwt_authentication(req: Request, handler: Callable) -> Response:
 
         # JWK and JWTClaims parameters for decoding
         key = environ.get('PUBLIC_KEY', None)
-        # TODO: get Oauth2 public key
-        # TODO: verify audience claim
+
+        # Get Oauth2 public key and verify audience claim here
```
otahontas, line 77: I'd delete these "here should be something" -lines totally. If you want to mark the lines where something should happen, link lines to issue #96

metadata_backend/api/middlewares.py
```diff
@@ -73,15 +73,15 @@ async def jwt_authentication(req: Request, handler: Callable) -> Response:
...
 
         # Include claims that are required to be present
         # in the payload of the token
         claims_options = {
             "iss": {
                 "essential": True,
-                # TODO: add the proper issuer URLs here
+                # Proper issuer URLs need to be added as the values
```
otahontas, line 84: remove this

metadata_backend/api/middlewares.py
```diff
@@ -95,7 +95,7 @@ async def jwt_authentication(req: Request, handler: Callable) -> Response:
             claims.validate()
             LOG.info('Auth token decoded and validated.')
 
-            # TODO: Retrieve GA4GH passports and process into permissions
+            # Retrieve GA4GH passports and process into permissions here
```
otahontas, line 98: this too

metadata_backend/api/middlewares.py
csc-jm: Fair enough :D

csc-jm:
Issues #96 and #98 are follow-up issues to this and mention the lines that previously had the TODO comments.

metadata_backend/api/middlewares.py
```diff
@@ -1,53 +1,113 @@
...
-    return http_error_handler
+        # Check token has proper scheme and was provided.
+        if not re.match('Bearer', scheme):
+            raise web.HTTPUnauthorized(reason="Invalid token scheme, "
+                                              "Bearer required.")
+        if token is None:
+            raise web.HTTPUnauthorized(reason='Token cannot be empty.')
+
+        # JWK and JWTClaims parameters for decoding
+        key = environ.get('PUBLIC_KEY', None)
```
blankdots, line 75: we resolve to getting a public key locally when we cannot reach a JWK endpoint, but usually we use a JWK endpoint to get the public key.

Issue #21, opened 2020-05-27
Add possibility to create PUT and PATCH requests to object recourse uris


otahontas:
#### Proposed solution
It should be possible to update metadata objects fully or partially. Make it possible to update objects via /objects/<schema>/<accessionId> and via /<schema/<accessionId> (redirection). 

#### DoD (Definition of Done)
- Objects can be updates via endpoints

#### Testing
- Test with unit tests

otahontas:
We need to think, whether this should be actually possible at all. 

Some rproblems that might happen:
- XMLs can't be just partially updated in database (that's hard)
- if user updates JSON that has been generated from XML, JSON and XML are not in-sync. How to prevent this?

blankdots:
partial (`patch`) upload should not be possible with XML, we will only do that to JSON
replace (`put`) I think should be the same, however I do see the need for it in some cases, where the XML has typo and someone corrects it.

PR #99, opened 2020-07-26
Add draft submission feature

### LINKED ISSUE #53: Add endpoint for saving draft versions of submissions
#### Proposed solution

We need possibility to (auto)save drafts from frontend before user decides to do the final submissions. An endpoint which saves forms as json to db and returns mongodb's internal ID should be enough first.

#### DoD (Definition of Done)
- Restful endpoint, maybe /objects/drafts/ which supports POST, GET and PUT 

#### Testing
- Unit tests

blankdots:
### Related issues

Fixes #53 

### Type of change

### Changes Made

1. added new endpoints for `drafts` submission
   * structuring into `/objects/drafts` proved challenging to use `/objects/{schema}/{accession_id}` therefore moved to separate endpoints
   * reused existing function
2. moved query code private function to address code reuse
3. add put endpoint for draft submission but making it general to address other object types
4. typo fixes in method documentation
5. raise error if deleted document not found also on update and replace
6. fix tests
7. other small fixes

### Testing

- [x] Unit Tests
- [x] Integration Tests

### DIFF
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -199,6 +199,30 @@ class RESTApiHandler:
                  f"in schema {collection} was successful.")
         return web.Response(status=204)
 
+    async def put_object(self, req: Request) -> Response:
+        """Save metadata object to database.
+
+        :param req: PUT request
+        :returns: JSON response containing accessionId for submitted object
+        """
+        schema_type = req.match_info['schema']
+        accession_id = req.match_info['accessionId']
+        self._check_schema_exists(schema_type)
+        collection = (f"draft-{schema_type}" if req.path.startswith("/drafts")
+                      else schema_type)
+
+        db_client = req.app['db_client']
+        content = await req.json()
+        operator = Operator(db_client)
+        await operator.replace_metadata_object(collection,
+                                               accession_id,
+                                               content)
+        body = json.dumps({"accessionId": accession_id})
+        LOG.info(f"PUT object with accesssion ID {accession_id} "
+                 f"in schema {collection} was successful.")
+        return web.Response(body=body, status=201,
+                            content_type="application/json")
+
 
 class SubmissionAPIHandler:
     """Handler for non-rest API methods."""
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -79,16 +79,19 @@ class RESTApiHandler:
         accession_id = req.match_info['accessionId']
         schema_type = req.match_info['schema']
         self._check_schema_exists(schema_type)
+        collection = (f"draft-{schema_type}" if req.path.startswith("/drafts")
+                      else schema_type)
 
         format = req.query.get("format", "json").lower()
         db_client = req.app['db_client']
-        operator = (XMLOperator(db_client) if format == "xml"
+        operator = (XMLOperator(db_client) if (format == "xml" and
+                    not req.path.startswith("/drafts"))
                     else Operator(db_client))
-        data, content_type = await operator.read_metadata_object(schema_type,
+        data, content_type = await operator.read_metadata_object(collection,
                                                                  accession_id)
         data = data if format == "xml" else json.dumps(data)
-        LOG.info(f"GET object with accesssion ID {accession_id}"
-                 f"from schema {schema_type}.")
+        LOG.info(f"GET object with accesssion ID {accession_id} "
+                 f"from schema {collection}.")
         return web.Response(body=data, status=200, content_type=content_type)
 
     async def post_object(self, req: Request) -> Response:
@@ -99,21 +102,24 @@ class RESTApiHandler:
         """
         schema_type = req.match_info['schema']
         self._check_schema_exists(schema_type)
+        collection = (f"draft-{schema_type}" if req.path.startswith("/drafts")
+                      else schema_type)
 
         db_client = req.app['db_client']
         operator: Union[Operator, XMLOperator]
-        if req.content_type == "multipart/form-data":
+        if (req.content_type == "multipart/form-data" and
+           not req.path.startswith("/drafts")):
             files = await _extract_xml_upload(req, extract_one=True)
             content, _ = files[0]
             operator = XMLOperator(db_client)
         else:
             content = await req.json()
             operator = Operator(db_client)
-        accession_id = await operator.create_metadata_object(schema_type,
+        accession_id = await operator.create_metadata_object(collection,
                                                              content)
         body = json.dumps({"accessionId": accession_id})
-        LOG.info(f"POST object with accesssion ID {accession_id}"
-                 f"in schema {schema_type} was successful.")
+        LOG.info(f"POST object with accesssion ID {accession_id} "
+                 f"in schema {collection} was successful.")
         return web.Response(body=body, status=201,
                             content_type="application/json")
 
@@ -125,11 +131,7 @@ class RESTApiHandler:
         """
         schema_type = req.match_info['schema']
         self._check_schema_exists(schema_type)
-
-        format = req.query.get("format", "json").lower()
-        if format == "xml":
-            reason = "xml-formatted query results are not supported"
-            raise web.HTTPBadRequest(reason=reason)
+        return await self._handle_query(req)
 
         def get_page_param(param_name: str, default: int) -> int:
             """Handle page parameter value extracting."""
@@ -175,13 +177,15 @@ class RESTApiHandler:
         """
         schema_type = req.match_info['schema']
         self._check_schema_exists(schema_type)
+        collection = (f"draft-{schema_type}" if req.path.startswith("/drafts")
+                      else schema_type)
 
         accession_id = req.match_info['accessionId']
         db_client = req.app['db_client']
-        await Operator(db_client).delete_metadata_object(schema_type,
+        await Operator(db_client).delete_metadata_object(collection,
                                                          accession_id)
-        LOG.info(f"DELETE object with accesssion ID {accession_id}"
-                 f"in schema {schema_type} was successful.")
+        LOG.info(f"DELETE object with accesssion ID {accession_id} "
+                 f"in schema {collection} was successful.")
         return web.Response(status=204)
 
 
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -39,10 +39,7 @@ class RESTApiHandler:
         :param req: GET request with query parameters
         :returns: JSON with query results
         """
-        schema_type = req.match_info['schema']
-        collection = (f"draft-{schema_type}" if req.path.startswith("/drafts")
-                      else schema_type)
-
+        collection = req.match_info['schema']
         format = req.query.get("format", "json").lower()
         if format == "xml":
             reason = "xml-formatted query results are not supported"
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -23,10 +23,10 @@ class RESTApiHandler:
     """Handler for REST API methods."""
 
     def _check_schema_exists(self, schema_type: str) -> None:
-        """Check is schema type exists.
+        """Check if schema type exists.
 
         :param schema_type: schema type.
-        :returns: None or HTTPNotFound if schema does not exist.
+        :raises: HTTPNotFound if schema does not exist.
         """
         if schema_type not in schema_types.keys():
             reason = f"Specified schema {schema_type} was not found."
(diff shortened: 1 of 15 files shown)


metadata_backend/api/handlers.py
```diff
@@ -171,18 +188,42 @@ async def delete_object(self, req: Request) -> Response:
...
+        operator = Operator(db_client)
+        await operator.replace_metadata_object(collection,
+                                               accession_id,
+                                               content)
+        body = json.dumps({"accessionId": accession_id})
+        LOG.info(f"PUT object with accesssion ID {accession_id} "
+                 f"in schema {collection} was successful.")
+        return web.Response(body=body, status=201,
+                            content_type="application/json")
+
```
csc-jm, line 226: I suppose this partially fixes issue #21 as well

metadata_backend/api/handlers.py
```diff
@@ -22,6 +22,64 @@
 class RESTApiHandler:
     """Handler for REST API methods."""
 
+    def _check_schema_exists(self, schema_type: str) -> None:
+        """Check is schema type exists.
```
csc-jm, line 26:
```suggestion
        """Check if schema type exists.
```

metadata_backend/api/handlers.py
```diff
@@ -22,6 +22,64 @@
...
+
+        :param schema_type: schema type.
+        :returns: None or HTTPNotFound if schema does not exist.
```
csc-jm, line 29:
```suggestion
        :raises: HTTPNotFound if schema does not exist.
```

metadata_backend/api/handlers.py
```diff
@@ -22,6 +22,64 @@
...
+        """
+        if schema_type not in schema_types.keys():
+            reason = f"Specified schema {schema_type} was not found."
+            LOG.error(reason)
+            raise web.HTTPNotFound(reason=reason)
+
+    async def _handle_query(self, req: Request) -> Response:
+        """."""
```
csc-jm, line 37:
```suggestion
        """Handle query results.
        
        :param req: GET request with query parameters
        :returns: JSON with query results
        """
```

metadata_backend/api/handlers.py
blankdots: I can add a patch endpoint as well, the `update` is not that difficult but maybe different PR this one is already pretty big

blankdots: I will assign that to myself so I don't forget and will add a follow-up PR after this one is merged

metadata_backend/api/operators.py
```diff
@@ -336,6 +421,21 @@ async def _format_data_to_create_and_add_to_db(self, schema_type: str,
...
+
+        XML is validated, then parsed to json and json is added to database.
+        After successful json insertion, xml itself is backed up to database.
+
+        :param schema_type: Schema type of the object to replace.
+        :param accession_id: Identifier of object to replace.
+        :param data: Original xml content
+        :returns: Accession Id for object inserted to database
+        """
+        raise web.HTTPNotImplemented
```
otahontas, line 437: Why do we add this if we're not implementing it?

metadata_backend/database/db_service.py
```diff
@@ -95,29 +96,46 @@ async def update(self, collection: str, accession_id: str,
...
-                                                            update_operation)
-        LOG.debug(f"DB doc updated for {accession_id}.")
-        return result.acknowledged
+        find_by_id = {"accessionId": accession_id}
+        update_op = {"$set": data_to_be_updated}
+        old_data = await self.database[collection].find_one(find_by_id)
+        if not old_data:
+            reason = f"Object with accession id {accession_id} was not found."
+            LOG.error(reason)
+            raise web.HTTPNotFound(reason=reason)
```
otahontas, line 105: For consistency, I would say its good to have all aiohttp related logic in operators module, and keep this module purely db related. Maybe return something thats telling to raise errors on operators?

otahontas: Open for opposing views as well :)

metadata_backend/database/db_service.py
```diff
@@ -95,29 +96,46 @@ async def update(self, collection: str, accession_id: str,
...
-        result = await self.database[collection].replace_one(find_by_id_query,
-                                                             new_data)
-        LOG.debug(f"DB doc replaced for {accession_id}.")
-        return result.acknowledged
+        find_by_id = {"accessionId": accession_id}
+        old_data = await self.database[collection].find_one(find_by_id)
+        if not old_data:
+            reason = f"Object with accession id {accession_id} was not found."
+            LOG.error(reason)
+            raise web.HTTPNotFound(reason=reason)
```
otahontas, line 130: Same consistency argument applies here

metadata_backend/database/db_service.py
```diff
@@ -127,10 +145,15 @@ async def delete(self, collection: str, accession_id: str) -> None:
...
-        find_by_id_query = {"accessionId": accession_id}
-        result = await self.database[collection].delete_one(find_by_id_query)
-        LOG.debug(f"DB doc deleted for {accession_id}.")
-        return result.acknowledged
+        find_by_id = {"accessionId": accession_id}
+        result = await self.database[collection].delete_one(find_by_id)
+        if result.deleted_count < 1:
+            reason = f"Object with accession id {accession_id} was not found."
+            LOG.error(reason)
+            raise web.HTTPNotFound(reason=reason)
```
otahontas, line 153: Same consistency argument applies here

tests/integration/run_tests.py
```diff
@@ -48,6 +57,19 @@ async def create_request_data(schema, filename):
...
 
+async def create_request_json_data(schema, filename):
+    """Create request data from pairs of schemas and filenames.
+
+    :param schema: name of the schema (folder) used for testing
+    :param filename: name of the file used for testing.
+    """
+    path_to_file = testfiles_root / schema / filename
+    path = path_to_file.as_posix()
+    with open(path, mode='r') as f:
```
otahontas, line 68: either use `aiofiles` here or change the other create_request_data function to be blocking as well

tests/test_db_service.py
```diff
@@ -95,13 +99,45 @@ async def test_replace_replaces_data(self):
...
         self.collection.delete_one.return_value = futurized(
-            DeleteResult({}, True)
+            DeleteResult({'n': 1}, True)
         )
         success = await self.test_service.delete("test", self.id_stub)
         self.collection.delete_one.assert_called_once_with({"accessionId":
                                                            self.id_stub})
         self.assertTrue(success)
 
+    async def test_update_raises(self):
```
otahontas, line 109: following tests should be moved if raising is moved to operators

metadata_backend/api/operators.py
csc-jm: `_format_data_to_replace_and_add_to_db()` was set as an abstract method so I'm guessing this is here so that the syntax is not broken. If the idea is to implement this later, then I'm fine with it being like this for now.

metadata_backend/database/db_service.py
blankdots: we need to raise an error for data not found, what error would be raised here if not one for  Http Not found ?

metadata_backend/api/operators.py
blankdots: @csc-jm is right  also we will use that when implementing XML file type replace in #21

metadata_backend/api/handlers.py
```diff
@@ -22,6 +22,64 @@
...
+        schema_type = req.match_info['schema']
+        collection = (f"draft-{schema_type}" if req.path.startswith("/drafts")
```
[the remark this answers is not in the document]
blankdots: ok, so the purpose is just to have CRUD on drafts no browsing them, i thought that was the case from the issue

metadata_backend/api/handlers.py
```diff
@@ -22,6 +22,64 @@
...
+
+        :param schema_type: schema type.
+        :returns: None or HTTPNotFound if schema does not exist.
+        """
+        if schema_type not in schema_types.keys():
+            reason = f"Specified schema {schema_type} was not found."
+            LOG.error(reason)
+            raise web.HTTPNotFound(reason=reason)
+
+    async def _handle_query(self, req: Request) -> Response:
```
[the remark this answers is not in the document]
blankdots: felt that it was a bit complicated that is why the split

metadata_backend/api/handlers.py
```diff
@@ -171,18 +188,42 @@ async def delete_object(self, req: Request) -> Response:
...
-        await Operator(db_client).delete_metadata_object(schema_type,
+        await Operator(db_client).delete_metadata_object(collection,
                                                          accession_id)
-        LOG.info(f"DELETE object with accesssion ID {accession_id}"
-                 f"in schema {schema_type} was successful.")
+        LOG.info(f"DELETE object with accesssion ID {accession_id} "
+                 f"in schema {collection} was successful.")
         return web.Response(status=204)
 
+    async def put_object(self, req: Request) -> Response:
```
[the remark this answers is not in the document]
blankdots: already have code for that it will follow this PR.

metadata_backend/database/db_service.py
blankdots: did not see such error in `motor`. I assume in can raise a general one/ custom one and then catch it in operators, but feels like a lot of code when in can just raise it here .

otahontas: I mean that all the other http errors are raised inside api-packages modules (handlers, operators, middlewares). We're breaking that logic by raising errors here in db_service module

metadata_backend/api/operators.py
otahontas: ok, no problem, I was just wondering why to add it before implementing.

metadata_backend/database/db_service.py
blankdots: ok, i see then we need a function in db to check there is data and based on that raise an error on operators before doing anything ... ?

metadata_backend/api/handlers.py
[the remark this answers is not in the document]
otahontas: I'd imagine we don't need browsing for drafts, since drafts are user-based can be listed in /users/{user} document when we implement that?

metadata_backend/api/handlers.py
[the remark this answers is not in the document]
otahontas: that's cool!

metadata_backend/database/db_service.py
blankdots: i was thinking this was easy to change, but it involves more logic to change as the operator is expecting true / false / None and errors with HTTPBadRequest if it gets anything else than true, which is not correct - what I am saying is either we raise or return something it is still a trade off to make.

if i implement a function to check data is there, while it works for replace, update, then comes the use of `deleted_count` which I can make use of and not do another db call to find if data exists.

anyway will try to figure something without a lot of changes

tests/integration/run_tests.py
blankdots: `aiofiles` was not meant to work nicely for simple json loading ... will do my best but just saying.

metadata_backend/database/db_service.py
otahontas: If my suggestion is bigger task to do / overcomplicating things, we can do this like you originally did

otahontas: I mean, let's not make things harder on purpose

tests/integration/run_tests.py
otahontas: my point was just that it's weird if we use aiofiles for xml loading and standard lib for json loading :D I suggests we remove aiofiles

metadata_backend/api/handlers.py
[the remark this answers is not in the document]
blankdots: removed.

metadata_backend/database/db_service.py
blankdots: refactored, hope now it is ok.

metadata_backend/database/db_service.py
blankdots: refactored

metadata_backend/database/db_service.py
blankdots: refactored

tests/test_db_service.py
blankdots: refactored and done.

tests/integration/run_tests.py
blankdots: i kept it, figured out what it needed.

Issue #97, opened 2020-07-21
Add support for adding metadata object folders 


otahontas:
#### Proposed solution

Users should be able to collect metadata objects into folders. Folders are mainly views to how objects are linked together, so at first we could have following:
- endpoint with crud methods for folder: 
   - /folders/ GET --> return all folders, POST --> add new
   - /folders/{folder_id} GET --> return one folder, PUT --> replace one folder, PATCH --> partially update one folder, DELETE --> delete folder
- support for folders in database

Suggestion for folder-json structure: 
```
{
  id: "Mongodb generated Id", 
  metadata_objects: [
{ "accessionId": "EGA12356", 
  "schema": "Study" }, 
{ "accessionId": "EGA12356339", 
  "schema": "Sample" }, 
{ "accessionId": "EGA12356398", 
  "schema": "Experiment" }, 
]
}
```

#### DoD (Definition of Done)
Folder endpoints are implemented 

#### Testing
Unit and integration tests

otahontas:
This is a big issue, so feel free to break it to smaller pieces

blankdots:
We can close the issue as most things were addressed by the PRs.
if need be we can create additional ones.

PR #101, opened 2020-07-27
Add folder creation endpoint

csc-jm:
### Description

First PR dealing with object folder API methods.

### Related issues

Part of #97 

### Type of change

### Changes Made

- Adds routes for the endpoint
- Adds object folder methods
- Implements a method for creating a new folder and adding it to db
- Some typo fixes spotted along the way

### Testing

- [x] Unit Tests, albeit better testing will be needed

### DIFF
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -182,7 +182,7 @@ class RESTApiHandler:
     async def delete_object(self, req: Request) -> Response:
         """Delete metadata object from database.
 
-        :param req: POST request
+        :param req: DELETE request
         :returns: JSON response containing accessionId for submitted object
         """
         schema_type = req.match_info['schema']
@@ -227,6 +227,53 @@ class RESTApiHandler:
                  f"in schema {collection} was successful.")
         return web.Response(body=body, status=201,
                             content_type="application/json")
+    async def get_folders(self, req: Request) -> Response:
+        """Get all possible object folders from database.
+
+        :param req: GET Request
+        :returns: JSON list of folders
+        """
+        raise NotImplementedError
+
+    async def post_folder(self, req: Request) -> Response:
+        """Save object folder to database.
+
+        :param req: POST request
+        :returns: JSON response containing folderId for submitted object
+        """
+        raise NotImplementedError
+
+    async def get_folder(self, req: Request) -> Response:
+        """Get one object folder by its folder id.
+
+        :param req: GET request
+        :returns: JSON response containing object folder
+        """
+        raise NotImplementedError
+
+    async def replace_folder(self, req: Request) -> Response:
+        """Replace object folder with a specific folder id.
+
+        :param req: PUT request
+        :returns: TBD
+        """
+        raise NotImplementedError
+
+    async def update_folder(self, req: Request) -> Response:
+        """Update object folder with a specific folder id.
+
+        :param req: PATCH request
+        :returns: TBD
+        """
+        raise NotImplementedError
+
+    async def delete_folder(self, req: Request) -> Response:
+        """Delete object folder from database.
+
+        :param req: DELETE request
+        :returns: TBD
+        """
+        raise NotImplementedError
 
     async def patch_object(self, req: Request) -> Response:
         """Update metadata object in database.
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -333,6 +333,7 @@ class RESTApiHandler:
                  f"in schema {collection} was successful.")
         return web.Response(body=body, status=201,
                             content_type="application/json")
+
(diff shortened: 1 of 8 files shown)


metadata_backend/api/handlers.py
```diff
@@ -158,6 +158,65 @@ async def delete_object(self, req: Request) -> Response:
...
+        raise web.HTTPNotImplemented
+
+    async def post_folder(self, req: Request) -> Response:
+        """Save object folder to database.
+
+        :param req: POST request
+        :returns: JSON response containing folder ID for submitted object
+        """
+        db_client = req.app['db_client']
+        operator = Operator(db_client)
```
otahontas, line 176: I don't think folder operations need to be proxyed through operator-class. Or at least then you should create some kind of FolderOperator, since the current operator class logic is done for xml/json metadata objects.

It's also more logical if folders are in database called "folders", so my suggestion would be to add here: `db_service = DBservice("folders", db_client)` and do db operations with it.

metadata_backend/api/operators.py
```diff
@@ -96,9 +96,21 @@ async def delete_metadata_object(self, schema_type: str,
         await self._remove_object_from_db(XMLOperator(db_client),
                                           schema_type,
                                           accession_id)
-        LOG.info(f"removing object with schema {schema_type} from database "
+        LOG.info(f"Removing object with schema {schema_type} from database "
                  f"and accession id: {accession_id}")
 
+    async def create_object_folder(self, data: Dict) -> str:
```
otahontas, line 102: not needed, if we do folder creation directly from handlers

metadata_backend/api/operators.py
```diff
@@ -138,6 +150,26 @@ async def _remove_object_from_db(self,
...
             raise web.HTTPBadRequest(reason=reason)
 
+    async def _insert_new_folder_to_db(self, data: Dict) -> str:
+        """Insert new object folder to database.
+
+        :param data: JSON object containing initial folder info
+        :returns: Folder ID for folder inserted to database
+        """
+        try:
+            data['metadata_objects'] = []
```
otahontas, line 160: I don't think there's need to add empty list? User can see there's no objects, if there's only name and description available

metadata_backend/api/operators.py
```diff
@@ -138,6 +150,26 @@ async def _remove_object_from_db(self,
...
+            insert = await self.db_service.create_folder(data)
+        except (ConnectionFailure, OperationFailure) as error:
+            reason = f"Error happened while getting file: {error}"
+            LOG.error(reason)
+            raise web.HTTPBadRequest(reason=reason)
+        if insert[0]:
```
otahontas, line 166: this is somewhat hard to read

metadata_backend/database/db_service.py
```diff
@@ -155,3 +155,15 @@ async def get_count(self, collection: str, query: Dict) -> int:
         """
         LOG.debug("DB doc count performed.")
         return await self.database[collection].count_documents(query)
+
+    @auto_reconnect
+    async def create_folder(self, folder_data: Dict) -> Tuple[bool, str]:
```
otahontas, line 160: This is almost the same code than in `create` method below, maybe just modify it a bit, e.g.
```
async def create(self, collection: str, document: Dict, return_folder_id: bool = False) -> Union[bool, string]:
...
if return_folder_id:
    return return.inserted_id
```

Also, you don't need to check if result was acknowledged. If id is returned, operation was successful, if None / error was raised, it wasn't. 

Acknowledgion is checked in metadata object methods, because we don't use the mongodb internal id in our responses (however, I'm not even sure if that acknowledgion check is the best way to do that...).

metadata_backend/api/operators.py
otahontas: if you implement create / create_folder (however you want to go) to return id , you can just do following: 
```
id = await self.db_service.create("folder", data)
...
if id: 
   return id
else: 
...
```

metadata_backend/api/handlers.py
blankdots: second this, i think it should be in different db collection

metadata_backend/database/db_service.py
```diff
@@ -61,16 +61,20 @@ def __init__(self, database_name: str,
...
         :param collection: Collection where document should be inserted
         :param document: Document to be inserted
-        :returns: True if operation was successful
+        :returns: True if operation was successful or ID of the created folder
         """
         result = await self.database[collection].insert_one(document)
         LOG.debug("DB doc inserted.")
-        return result.acknowledged
+        if isFolder:
+            return result.inserted_id
```
blankdots, line 75: maybe we can generate something else, to skip relying on internal DB IDs and issues that come with exposing internal ids see `_generate_accession_id` in `operators.py`

PR #104, opened 2020-07-30
Add folder get methods

csc-jm:
### Description

So this PR includes the get methods for a single folder and all folders. Proper testing still lacks but I currently believe things here work more or less.

### Related issues

Part of #97 

### Type of change

### Changes Made

- moved `patch_object()` method up so it's not below all the folder methods
- added `get_folders()` method
- added `get_folder()` method
- edited `db_service.read()` to accommodate folders
- added couple unit tests 
- fixed isort errors

### Testing

- [x] Unit tests

### Mentions
POST content check implemented the same way as @blankdots in #107

### DIFF
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -302,12 +302,6 @@ class RESTApiHandler:
         """
         db_client = req.app['db_client']
         db_service = DBService("folders", db_client)
-        """
-        if not db_service.database["folder"].count_documents({}):
-            reason = f"No folders found."
-            LOG.error(reason)
-            raise web.HTTPNotFound(reason=reason)
-        """
         cursor = db_service.query("folder", {})
         folders = [folder async for folder in cursor]
         body = json.dumps({"folders": folders})
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -301,7 +301,13 @@ class RESTApiHandler:
         :returns: JSON list of folders available for the user
         """
         db_client = req.app['db_client']
-        db_service = DBService('folders', db_client)
+        db_service = DBService("folders", db_client)
+        """
+        if not db_service.database["folder"].count_documents({}):
+            reason = f"No folders found."
+            LOG.error(reason)
+            raise web.HTTPNotFound(reason=reason)
+        """
         cursor = db_service.query("folder", {})
         folders = [folder async for folder in cursor]
         body = json.dumps({"folders": folders})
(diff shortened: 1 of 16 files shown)


metadata_backend/api/handlers.py
```diff
@@ -262,13 +262,54 @@ async def put_object(self, req: Request) -> Response:
...
         """Get all possible object folders from database.
 
         :param req: GET Request
-        :returns: JSON list of folders
+        :returns: JSON list of folders available for the user
         """
-        raise web.HTTPNotImplemented
+        db_client = req.app['db_client']
+        db_service = DBService('folders', db_client)
+        collection = db_service.database['folder']
```
otahontas, line 304: We can use query method from db_services here and then collect folders with list comprehension:

```
cursor = db_service.query("folder", {})
folders = [folder async for folder in cursor]
```

metadata_backend/api/handlers.py
```diff
@@ -262,13 +262,54 @@ async def put_object(self, req: Request) -> Response:
...
+            operator = Operator(db_client)
+        await operator.update_metadata_object(collection,
+                                              accession_id,
+                                              content)
+        body = json.dumps({"accessionId": accession_id})
+        LOG.info(f"PUT object with accesssion ID {accession_id} "
+                 f"in schema {collection} was successful.")
+        return web.Response(body=body, status=201,
+                            content_type="application/json")
+
```
otahontas, line 295: Add `@auto_reconnect` decorator here so query db operation is reconnected when needed

metadata_backend/api/handlers.py
```diff
@@ -15,12 +15,12 @@
 
 from ..conf.conf import schema_types
 from ..database.db_service import DBService
```
otahontas, line 17: `from ..database.db_service import DBService, auto_reconnect`

metadata_backend/database/db_service.py
```diff
@@ -86,15 +86,18 @@ async def exists(self, collection: str, accession_id: str) -> bool:
...
         """Find object by its accessionId.
 
         :param collection: Collection where document should be searched from
-        :param accession_id: Accession id of the document to be searched
+        :param id: Accession id of the document or folder id of the folder
         :returns: First document matching the accession_id
         """
-        find_by_id = {"accessionId": accession_id}
-        LOG.debug(f"DB doc read for {accession_id}.")
+        if collection == "folder":
```
otahontas, line 96: maybe 
```
if collection == "folder" id_key = "folderId" else "accessionId"
find_by_id = {id_key: id}
```

tests/test_handlers.py
```diff
@@ -434,3 +434,21 @@ async def test_folder_creation_works(self):
         self.MockedDbService().create.assert_called_once()
         self.assertEqual(response.status, 201)
         self.assertEqual(json_resp['folderId'], self.folder_id)
+
+    @unittest_run_loop
+    async def test_get_folders(self):
+        """Test get_folders() endpoint returns empty list."""
+        self.MockedDbService().db_client = MagicMock()
```
otahontas, line 441: this doesn't actually test much. If you want to test get_folders, I'd suggest you to mock `db_service.query` to return avaitable cursor, probably with the help from MockCursor class in test_operators.

metadata_backend/api/handlers.py
csc-jm: this was a nice fix 👍

metadata_backend/api/handlers.py
csc-jm: done

metadata_backend/database/db_service.py
csc-jm: done

tests/test_handlers.py
csc-jm: this was also very helpful 👍

metadata_backend/api/handlers.py
```diff
@@ -303,8 +351,24 @@ async def get_folder(self, req: Request) -> Response:
...
+            if not folder:
+                reason = f"Folder with id {folder_id} not found."
+                LOG.error(reason)
+                raise web.HTTPNotFound(reason=reason)
+        except (ConnectionFailure, OperationFailure) as error:
+            reason = f"Error happened while getting folder: {error}"
+            LOG.error(reason)
+            raise web.HTTPBadRequest(reason=reason)
+
+        folder.pop('_id', None)  # remove unneccessary mongodb id from result
```
otahontas, line 368: btw you can just do `del doc["_id"]` here, since you don't use the value returned by `.pop()`.

blankdots: `del` will raise `KeyError` if it does not find the key, I think this is ok. Although it might never be raised it is better to avoid and fail silently as it is not a relevant key.

otahontas: well, but mongoDb will always return _id, right? if you think failing silently is better, this should be also fixed: https://github.com/CSCfi/metadata-submitter/blob/2de12d9f26837025a172eb1d60ea5d42d713351a/metadata_backend/api/operators.py#L497

blankdots: maybe we can prevent it on query level, i think we can merge as is.

PR #107, opened 2020-07-31
Json validation

### LINKED ISSUE #86: Add json schema to validation to backend
#### Proposed solution
Validate JSON submissions POSTed to `/objects/{schema}` against json schemas. Let's use https://github.com/dchaplinsky/aiohttp_validate

#### DoD (Definition of Done)
Json is validated in the same manner as XMLs, invalid submissions are reported.

#### Testing
Unit tests and integration tests

### LINKED ISSUE #87: Validate JSON converted from XMLs with json schemas
#### Proposed solution
Validate JSONs converted from XMLs to see if conversion was successful.

#### DoD (Definition of Done)
JSONs are validated, errors are reported

#### Testing
Unit tests.

blankdots:
### Related issues

Fixes #86 
Fixes #87 

### Type of change

### Changes Made

1. fixed `filetype` in schemas so that it reflects that `files` can be both an object and and array
2. fixed test json files fixtures for correct validation in integration tests
3. added JSONValidator class and calls for it in `POST` and `PUT` Request
4. for `PATCH` we do a sanity check and validate the data changes are still valid in db
   - not ideal as we do it after the update
   - the reason to keep it like this is that `PATCH` is only done on draft and folders, not actual objects and we should not do `PATCH` on objects, that will open up for other issues (e.g. accidental overwrite of objects with invalid data).
5. validate XML2JSON conversion output
6. `isort` fixes :roll_eyes: 
7. fix unit tests

### Testing

- [x] Unit Tests
- [x] Integration Tests

### Mentions

@csc-jm some similar changes were done in #104 I am fine to wait for them

### DIFF
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -17,9 +17,9 @@ from ..conf.conf import schema_types
 from ..database.db_service import DBService, auto_reconnect
 from ..helpers.logger import LOG
 from ..helpers.parser import XMLToJSONParser
-from ..helpers.schema_loader import (JSONSchemaLoader, SchemaNotFoundException,
-                                     XMLSchemaLoader)
-from ..helpers.validator import XMLValidator
+from ..helpers.schema_loader import (XMLSchemaLoader, SchemaNotFoundException,
+                                     JSONSchemaLoader)
+from ..helpers.validator import XMLValidator, JSONValidator
 from .operators import Operator, XMLOperator
 
 
@@ -178,6 +178,9 @@ class RESTApiHandler:
     async def post_object(self, req: Request) -> Response:
         """Save metadata object to database.
 
+        For JSON request body we validate it is consistent with the
+        assosicated JSON schema.
+
         :param req: POST request
         :returns: JSON response containing accessionId for submitted object
         """
@@ -194,6 +197,7 @@ class RESTApiHandler:
             operator = XMLOperator(db_client)
         else:
             content = await req.json()
+            JSONValidator(content, schema_type).validate
             operator = Operator(db_client)
         accession_id = await operator.create_metadata_object(collection,
                                                              content)
@@ -235,6 +239,9 @@ class RESTApiHandler:
     async def put_object(self, req: Request) -> Response:
         """Replace metadata object in database.
 
+        For JSON request body we validate that it consists of all the
+        required fields before replacing in the DB.
+
         :param req: PUT request
         :returns: JSON response containing accessionId for submitted object
         """
@@ -252,6 +259,7 @@ class RESTApiHandler:
             operator = XMLOperator(db_client)
         else:
             content = await req.json()
+            JSONValidator(content, schema_type).validate
             operator = Operator(db_client)
         await operator.replace_metadata_object(collection,
                                                accession_id,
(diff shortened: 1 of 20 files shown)


metadata_backend/api/operators.py
```diff
@@ -214,14 +219,20 @@ async def _update_object_from_db(self, schema_type: str,
                           "was not found.")
                 LOG.error(reason)
                 raise web.HTTPNotFound(reason=reason)
-            replace_success = (await self.db_service.update(schema_type,
-                                                            accession_id,
-                                                            data))
+            update_success = (await self.db_service.update(schema_type,
```
otahontas, line 222: regarding this: "The reason to keep it like this is that PATCH is only done on draft and folders, not actual objects and we should not do PATCH on objects, that will open up for other issues (e.g. accidental overwrite of objects with invalid data)."

If we don't want to do patches for objects, shouldn't we just prevent it in handlers.py in the first place
https://github.com/CSCfi/metadata-submitter/blob/6afe4c4a866fbcbaa8d916f92678c4dfc429e688/metadata_backend/api/handlers.py#L296

metadata_backend/helpers/validator.py
```diff
@@ -76,3 +80,75 @@ def is_valid(self) -> bool:
         """Quick method for checking validation result."""
         resp = json.loads(self.resp_body)
         return resp['isValid']
+
+
+def extend_with_default(validator_class: Draft7Validator) -> Draft7Validator:
```
otahontas, line 85: why do we do this, I thought we don't want to use default values in jsonschemas?

metadata_backend/api/handlers.py
```diff
@@ -193,7 +196,14 @@ async def post_object(self, req: Request) -> Response:
...
         else:
-            content = await req.json()
+            try:
+                content = await req.json()
+            except json.decoder.JSONDecodeError as e:
+                reason = ("JSON is not correctly formatted."
+                          f" See: {e}")
+                LOG.error(reason)
+                raise web.HTTPBadRequest(reason=reason)
+            JSONValidator(content, schema_type).validate
```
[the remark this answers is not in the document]
blankdots: well, I worked on existing code. Re-factoring maybe after we do a release

metadata_backend/api/operators.py
blankdots: it is not in `server.py` that is enough to say Method not allowed

metadata_backend/helpers/validator.py
blankdots: in case they are added  I do extra or will be added at some point for any reason e.g. custom schemas.

metadata_backend/api/handlers.py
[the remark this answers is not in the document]
blankdots: also where would the HTTP request validation be done if not on the handler ? any suggestions ?
In parser it is done, as that is where the output and conversion from XML 2 JSON is done

Not sure where we would shuffle them around.

metadata_backend/api/operators.py
blankdots: but I can add a check there just to be extra, extra sure.

metadata_backend/api/handlers.py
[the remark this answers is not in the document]
csc-jm: Maybe all the requests could be validated via a middleware?

PR #108, opened 2020-07-31
Remove id from mongo read

blankdots:
### Description

no need to remove `_id` from dictionary as we could just ask mongo not to return it.

### Type of change

Not exactly a bug, but still a fix that we should not do this processing on the dictionary

### Changes Made

1. added `{'_id': False}` to `find_one`
2. remove `del` and `pop` for `_id`
3 fix unit tests

### Testing

- [x] Tests do not apply

but still need to pass

### Mentions

triggered by: https://github.com/CSCfi/metadata-submitter/pull/104#pullrequestreview-459095753


otahontas:
Fixed one more objectId test, should be now ok

PR #116, opened 2020-08-10
Bugfix/fix more schema details

otahontas:
### Description

This PR adds some consistency fixes to json schemas.

### Type of change


otahontas:
Added alternative suggestion for librarylayout dependency. This is easier to render in frontend since we can have fields for `nominalLength` and `nominalSdev` always there, but report errors only when paired is checked.

Please note that this approach doesn't work with our backend validator (tried with experiment json test file), but frontend reports errors correctly.

PR #120, opened 2020-08-12
Bugfix/small typo fixes

otahontas:
### Description

Couple of small bugfixes

### Type of change

### Changes Made
- Fix bug where folder id wasn't returned after succesful deletion, resulting logging of "None" instead of if
- FIx bug where validation was run for folder read from database instead of patched version of folder, making it possible to patch folders to non-valid state, thus failing validation on following patches

### DIFF
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -628,6 +628,8 @@ class FolderOperator:
             upd_content = patch.apply(folder)
             JSONValidator(upd_content, "folders").validate
             update_success = await self.db_service.update("folder", folder_id, upd_content)
+            sanity_check = await self.db_service.read("folder", folder_id)
+            JSONValidator(sanity_check, "folders").validate
         except (ConnectionFailure, OperationFailure) as error:
             reason = f"Error happened while getting folder: {error}"
             LOG.error(reason)
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -625,8 +625,8 @@ class FolderOperator:
         await self._check_folder_exists(self.db_service, folder_id)
         try:
             folder = await self.db_service.read("folder", folder_id)
-            JSONValidator(folder, "folders").validate
             upd_content = patch.apply(folder)
+            JSONValidator(upd_content, "folders").validate
             update_success = await self.db_service.update("folder", folder_id, upd_content)
         except (ConnectionFailure, OperationFailure) as error:
             reason = f"Error happened while getting folder: {error}"
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -660,9 +660,13 @@ class FolderOperator:
             LOG.error(reason)
             raise web.HTTPBadRequest(reason=reason)
         if not delete_success:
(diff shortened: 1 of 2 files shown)


metadata_backend/api/operators.py
```diff
@@ -625,8 +625,8 @@ async def update_folder(self, folder_id: str, patch: JsonPatch) -> str:
         await self._check_folder_exists(self.db_service, folder_id)
         try:
             folder = await self.db_service.read("folder", folder_id)
-            JSONValidator(folder, "folders").validate
             upd_content = patch.apply(folder)
+            JSONValidator(upd_content, "folders").validate
```
csc-jm, line 629: This was meant to just be a sanity check so that the `folder` returned from `db_service.read()` is still fine. However, validating the patched content is probably pretty good to have. 👍

metadata_backend/api/operators.py
```diff
@@ -660,9 +660,12 @@ async def delete_folder(self, folder_id: str) -> None:
             LOG.error(reason)
             raise web.HTTPBadRequest(reason=reason)
         if not delete_success:
-            reason = "Deleting for {folder_id} from database failed."
+            reason = f"Deleting for {folder_id} from database failed."
             LOG.error(reason)
             raise web.HTTPBadRequest(reason=reason)
+        else:
+            LOG.info(f"Deleting folder with id {folder_id} to database succeeded.")
+            return folder_id
```
csc-jm, line 668: Ok, so this a pretty good catch and it wouldn't have passed in previous PR, had there been better testing done then, whoops 😅

However, I'm wondering if returning the `folder_id` is that necessary. The `delete_folder()` method in `handlers.py` could instead just be:
```
        folder_id = req.match_info["folderId"]
        db_client = req.app["db_client"]
        operator = FolderOperator(db_client)
        await operator.delete_folder(folder_id)
        LOG.info(f"DELETE folder with ID {folder_id} was successful.")
```

blankdots:
LGTM and we can merge.

metadata_backend/api/operators.py
blankdots: yeah this could respond with 204`, did not noticed it at first we should keep it consistent with other delete `endpoints`

metadata_backend/api/operators.py
otahontas: Well, I don't think sanity check is needed after first read where content hasn't been changed yet. But doing it after patching content in db is a good one, I'll add that

metadata_backend/api/operators.py
otahontas: good point, will fix

otahontas: i think it's better to actually catch result from awaited delete folder operation and then log. Otherwise you're logging that delete was succesful before it was actually executed

otahontas: so actually, I would argue no change is needed here

otahontas:
> i think delete can be of status `204`

it is already 204

metadata_backend/api/operators.py
csc-jm: alright, fair enough

PR #121, opened 2020-08-12
Add user endpoints

### LINKED ISSUE #105: Add user endpoint
#### Proposed solution
Add endpoints to save user data, especially folders created by user

#### DoD (Definition of Done)
User endpoint is there with POST, GET, PATCH, PUT (maybe DELETE if needed)

#### Testing
at least peer to peer, hopefully integration tests also

csc-jm:
### Description

Another endpoint PR without proper testing, even though original issue requested it. The PR somewhat effectively implements endpoints for user actions and probably works to an extent.

### Related issues

Fixes #105 

### Type of change

### Changes Made

- added `GET`, `PATCH` and `DELETE` endpoints for folders
- added `UserOperator` class
- edited `DBService` methods once again to account for users 
- added JSON schema for user objects
- some `black` formatting fixes and a typo fix or two here and there

### Testing

- [x] Unit Tests
- [x] Needs testing

### Mentions

Cheers to @blankdots for code style suggestion

### DIFF
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -370,23 +370,6 @@ class RESTApiHandler:
         LOG.info(f"DELETE folder with ID {folder} was successful.")
         return web.Response(status=204)
 
-    async def post_user(self, req: Request) -> Response:
-        """Save new user object to database.
-
-        :param req: POST request
-        :returns: JSON response containing user ID for new user
-        """
-        db_client = req.app["db_client"]
-        content = await self._get_data(req)
-        # JSONValidator(content, "users").validate
-        operator = UserOperator(db_client)
-        user = await operator.create_user(content)
-        body = json.dumps({"userId": user})
-        url = f"{req.scheme}://{req.host}{req.path}"
-        location_headers = {"Location": f"{url}/{user}"}
-        LOG.info(f"POST new user with ID {user} was successful.")
-        return web.Response(body=body, status=201, headers=location_headers, content_type="application/json")
-
     async def get_user(self, req: Request) -> Response:
         """Get one user by its user ID.
 
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -20,7 +20,7 @@ from ..helpers.schema_loader import (
     XMLSchemaLoader,
 )
 from ..helpers.validator import JSONValidator, XMLValidator
-from .operators import FolderOperator, Operator, XMLOperator
+from .operators import FolderOperator, Operator, XMLOperator, UserOperator
 
 
 class RESTApiHandler:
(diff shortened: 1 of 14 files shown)


metadata_backend/api/handlers.py
```diff
@@ -366,6 +370,64 @@ async def delete_folder(self, req: Request) -> Response:
...
 
+    async def post_user(self, req: Request) -> Response:
+        """Save new user object to database.
+
+        :param req: POST request
+        :returns: JSON response containing user ID for new user
+        """
+        db_client = req.app["db_client"]
+        content = await self._get_data(req)
+        # JSONValidator(content, "users").validate
```
blankdots, line 381: do we want it part of this PR?

metadata_backend/api/handlers.py
```diff
@@ -14,9 +14,13 @@
 from ..conf.conf import schema_types
 from ..helpers.logger import LOG
 from ..helpers.parser import XMLToJSONParser
-from ..helpers.schema_loader import JSONSchemaLoader, SchemaNotFoundException, XMLSchemaLoader
+from ..helpers.schema_loader import (
```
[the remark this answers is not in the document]
otahontas: this is probably result of first running black with 88 line length (default) and then 120. Maybe fix this back to how it was?

metadata_backend/api/handlers.py
```diff
@@ -366,6 +370,64 @@ async def delete_folder(self, req: Request) -> Response:
         LOG.info(f"DELETE folder with ID {folder} was successful.")
         return web.Response(status=204)
 
+    async def post_user(self, req: Request) -> Response:
```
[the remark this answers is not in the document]
otahontas: agreed

metadata_backend/api/handlers.py
```diff
@@ -366,6 +370,64 @@ async def delete_folder(self, req: Request) -> Response:
...
+    async def delete_user(self, req: Request) -> Response:
+        """Delete user from database.
+
+        :param req: DELETE request
+        :returns: HTTP No Content response
+        """
+        user_id = req.match_info["userId"]
+        db_client = req.app["db_client"]
+        operator = UserOperator(db_client)
+        await operator.delete_user(user_id)
```
otahontas, line 427: catch result, then return id

metadata_backend/api/operators.py
```diff
@@ -677,3 +677,119 @@ def _generate_folder_id(self) -> str:
...
+    async def update_user(self, user_id: str, data: JsonPatch) -> str:
+        """Update user object from database.
+
+        :param user_id: ID of user to update
+        :param data: Data to be updated
+        :returns: ID of the user updated to database
+        """
+        await self._check_user_exists(self.db_service, user_id)
+        try:
+            update_success = await self.db_service.update("user", user_id, data)
```
otahontas, line 746: I'd suggest you to write a json schema for users too and do following checks 

 1. Don't let patch to modify user data coming from external sources (probably name, id etc.), only folders and drafts for now.
2. Do validation for updated content like folders and optionally sanity check too after saving updated content to db

metadata_backend/api/operators.py
```diff
@@ -677,3 +677,119 @@ def _generate_folder_id(self) -> str:
...
+
+        if not update_success:
+            reason = "Updating user to database failed for some reason."
+            LOG.error(reason)
+            raise web.HTTPBadRequest(reason=reason)
+        else:
+            LOG.info(f"Updating user with id {user_id} to database " "succeeded.")
+            return user_id
+
+    async def delete_user(self, user_id: str) -> None:
```
otahontas, line 765: return id here

metadata_backend/api/handlers.py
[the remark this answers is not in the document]
csc-jm: Yea that's most likely the case. I prefer longer lines and less scrolling up and down so I changed them back even though it would be fine like that as well.

PR #129, opened 2020-08-19
Add more integration tests

### LINKED ISSUE #125: Add integration tests for folders and users
#### Proposed solution

Both folder and user endpoints were implemented with very minimal testing. Thus, integration tests would be needed.

#### DoD (Definition of Done)

Folders and users are tested properly

#### Testing

yes

csc-jm:
### Description

Integration tests for folders and users

### Related issues

Fixes #125 

### Type of change

### Changes Made

Adds integration tests

### Testing

- [x] Integration Tests

### DIFF
--- a/tests/integration/run_tests.py
+++ b/tests/integration/run_tests.py
@@ -381,19 +381,12 @@ async def test_crud_folders_works():
         folder_id = await patch_folder(sess, folder_id, patch)
         async with sess.get(f"{folders_url}/{folder_id}") as resp:
             LOG.debug(f"Checking that folder {folder_id} was patched")
-            folder = {
-                "folderId": folder_id,
-                "name": "test",
-                "description": "test folder",
-                "published": False,
-                "metadataObjects": [{"accessionId": accession_id, "schema": "sample"}],
-            }
             res = await resp.json()
-            assert res['folderId'] == folder_id, "content mismatch"
-            assert res['name'] == "test", "content mismatch"
-            assert res['description'] == "test folder", "content mismatch"
-            assert res['published'] == False, "content mismatch"
-            assert res['metadataObjects'] == [{"accessionId": accession_id, "schema": "sample"}], "content mismatch"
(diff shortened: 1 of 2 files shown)


tests/integration/run_tests.py
```diff
@@ -340,6 +448,15 @@ async def main():
...
+
+    # Test reading, updating and deleting folders
+    LOG.debug("=== Testing basic CRUD user operations ===")
+    await test_crud_users_works()
+
 
 if __name__ == "__main__":
-    asyncio.run(main())
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(main())
```
csc-jm, line 462: This because the original causes: `AttributeError: module 'asyncio' has no attribute 'run'` 
when I run this.

tests/integration/run_tests.py
```diff
@@ -317,6 +362,69 @@ async def test_getting_all_objects_from_schema_works():
...
+        async with sess.get(f"{folders_url}/{folder_id}") as resp:
+            LOG.debug(f"Checking that folder {folder_id} was patched")
+            folder = {
+                "folderId": folder_id,
+                "name": "test",
+                "description": "test folder",
+                "published": False,
+                "metadataObjects": [{accession_id, "sample"}],
+            }
+            assert resp.json == folder, "Response data error"
```
blankdots, line 388: i would assert individual values to be sure. See https://github.com/CSCfi/metadata-submitter/blob/master/tests/integration/run_tests.py#L202-L219

blankdots (changes requested):
something feels off with the checks in the user, or i did not understand it correctly,

tests/integration/run_tests.py
csc-jm: Good point

tests/integration/run_tests.py
csc-jm: Yea I realized this was the case but failed to properly understand how `pyenv` operates at the time. Now I do and I reverted back.

Issue #106, opened 2020-07-30
Add connections between users and folders


otahontas:
#### Proposed solution
For easy db queries, user docs should contain list of folders user has created and each folder should probably also contain info about the user who created the folder.

Also `get_folders` in handlers.py should return list of folders by user / somehow restrict users from browsing folders created by other people.

#### DoD (Definition of Done)
It's easy to query for users folders

#### Testing
peer to peer, integration, unit

blankdots:
> somehow restrict users from browsing folders created by other people

i think when processing any request we should see if user is authenticated and after that when checking a e.g. object resource check if the user has that resource in its objects, folders etc and then show it.

this task should come after: `/users` #105

csc-jm:
Regarding `get_folders` endpoint for user specific folders in `handlers.py`, a better way to achieve this would be to check the user logged in the session eg. from cookies. This functionality has not been implemented yet, so this issue can be delayed for now.

PR #133, opened 2020-10-05
Feature/oidc client integration

### LINKED ISSUE #98: Add proper authentication functionality
#### Proposed solution

On top of the authentication outlines set in the authentication middleware, couple things are lacking:
- OAuth2 server(s) should be determined that return a public key for JWT Bearer token validation
- audience claim that states the intended recipient of the token needs to be verified if one exists and is required
- authenticated bearer token issuers need to be included in payload claims

#### DoD (Definition of Done)

- public key is received from OAuth2 server(s)
- audience claim(s) are verified and included in `claims_options`
- authenticated bearer token issuers are referenced in `claims_options['iss']['values']` 

#### Testing

Possibility of integration testing

blankdots:
### Description

This would resolve authenticating via an OIDC endpoint. however to make use of this in a proper way #135 is needed
More testing will come after #135 and #106 

### Related issues

Closes #98 

### Type of change

### Changes Made

- Remove jwt auth as not needed anymore, used OIDC endpoint should be able to handle multiple auth types.
- added auth routes
- added middleware to check login when use authenticates.
- add configuration parameters.
- use 1 worker as session handling is problematic across workers

### Testing

- [x] Unit Tests

### Mentions

works with: https://github.com/CSCfi/metadata-submitter-frontend/pull/80


metadata_backend/conf/conf.py
```diff
@@ -89,3 +89,21 @@ def create_db_client() -> AsyncIOMotorClient:
...
+
+aai_config = {
+    "client_id": os.getenv("CSC_AAI_CLIENT_ID", "public"),
+    "client_secret": os.getenv("CSC_AAI_CLIENT_SECRET", "secret"),
+    "domain": os.getenv("BASE_URL", "http://localhost:5430"),
+    "scope": "openid profile email",
+    "iss": os.getenv("AUTH_URL", ""),
+    "aud": "aud1",
+    "callback_url": f'{os.getenv("BASE_URL", "http://localhost:5430")}/callback',
+    "auth_url": f'{os.getenv("OIDC_URL", "")}/authorize',
```
csc-jm, line 104: When I was testing this branch with the `feature/login-link` branch of the frontend repo, I ran into an error: `Oops! Page /authorize could not be found.`

I assume it's caused by this and that this should instead be:
```suggestion
    "auth_url": f'{os.getenv("OIDC_URL", "")}/aai',
```

csc-jm (changes requested):
Otherwise everything looks great apart from that one thing.

metadata_backend/conf/conf.py
blankdots: nope, that is correct. that is oidc endpoint configuration 
I suspect you were missing something from configuration in docker-compose or did not run the frontend at same url as backend. You need to build and start the full image not the dev one.
In the `Dockerfile` you need to `RUN git checkout feature/login-link` after line 9 to build the image with that branch

csc-jm (approved):
Great PR

PR #134, opened 2020-10-05
Feature/move to only37

blankdots:
### Description

Move to only 3.7 which will allow use to use move precise typing with Python.
These changes break 3.6 compatibility.

### Type of change

### Changes Made

- use CIMultiDict for headers
- remove use of 3.6

### Testing

- [x] Tests do not apply


csc-jm (approved):
I've been waited for this change 👍

PR #140, opened 2020-11-04
update and refactor gh actions

blankdots:
### Type of change

### Changes Made

1. update actions versions
2. better use of build docker image action

### Testing

- [x] Tests do not apply


blankdots:
Moving this further, as already tested and we can come back to fix it.

PR #142, opened 2020-11-10
Feature/csrf session

### LINKED ISSUE #136: check for csrf and do session check
#### Description

do better handling of cookies signature and checking if they correspond from same requester and even sign them

#### DoD (Definition of Done)

better handling of cookies and CSRF + session checking

#### Testing

peer review and unit test

blankdots:
### Related issues

fixes #136 

### Type of change

### Changes Made

1. improved workflow for AccessHandler to adhere to https://openid.net/specs/openid-connect-core-1_0.html also validates id token (reintroduce authlib)
2. use encrypted cookie to validate requests
3. handle csrf
4. track user session
5. other small fixes

### Testing

### DIFF
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -114,7 +114,7 @@ def _check_csrf(request: web.Request) -> bool:
     # site's wrong)
     if "Referer" in request.headers.keys():
         # Pass referer check if we're returning from the login.
-        if request.headers["Referer"] in aai_config["auth_url"]:
+        if request.headers["Referer"] in aai_config["referer"]:
             LOG.info("Skipping Referer check due to request coming from OIDC.")
             return True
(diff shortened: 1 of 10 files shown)


metadata_backend/api/middlewares.py
```diff
@@ -47,23 +50,98 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
...
+            "/schemas",
+            "/drafts",
+            "/validate",
+            "/submit",
+            "/folders",
+            "/objects",
+            "/users",
+            "/logout",
+            "/home",
+            "newdraft",
```
csc-jm, line 65: Should this instead be?
```suggestion
            "/newdraft",
```

blankdots: good catch

csc-jm (approved):
Other than that, I found no faults 👍

PR #144, opened 2020-11-16
Refactor/draft folders

csc-jm:
### Description

As discussed, a refactoring to the way drafts are stored in folders was in place. The current idea with the `publish_folder` endpoint in this PR is that the drafts in the `drafts` key of the folder that is to be published are removed from the database and the folder is simultaneously patched so that the drafts are removed from the folder.

### Type of change

- [x] Breaking change (or not exactly breaking but somewhat altering and requires a change to frontend)

### Changes Made

- added `drafts` key to the folders
- updated to the OpenApi specs and json schema
- created new endpoint for publishing a folder
- updated the existing integration test for folders
- other general style fixes 

### Testing

- [x] Unit Tests
- [x] Integration Tests

### DIFF
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -352,7 +352,7 @@ class RESTApiHandler:
 
         # Check patch operations in request are valid
         patch_ops = await self._get_data(req)
-        allowed_paths = ["/name", "/description", "/metadataObjects", "/published"]
+        allowed_paths = ["/name", "/description", "/metadataObjects", "/published", "/drafts"]
         for op in patch_ops:
             if all(i not in op["path"] for i in allowed_paths):
                 reason = f"Request contains '{op['path']}' key that cannot be updated to folders."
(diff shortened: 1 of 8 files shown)


metadata_backend/api/handlers.py
```diff
@@ -366,6 +366,37 @@ async def patch_folder(self, req: Request) -> Response:
         LOG.info(f"PATCH folder with ID {folder} was successful.")
         return web.Response(body=body, status=200, content_type="application/json")
 
+    async def publish_folder(self, req: Request) -> Response:
+        """Update object folder specifically into published state.
+
+        :param req: PATCH request
```
blankdots, line 372: this is a `patch` but in routes it looks like a `.get` .... maybe we can settle on something

metadata_backend/server.py
```diff
@@ -72,6 +72,7 @@ async def init() -> web.Application:
         web.get("/folders/{folderId}", rest_handler.get_folder),
         web.patch("/folders/{folderId}", rest_handler.patch_folder),
         web.delete("/folders/{folderId}", rest_handler.delete_folder),
+        web.get("/folders/publish/{folderId}", rest_handler.publish_folder),
```
blankdots, line 75: how about `/publish/{folderId}` ?

blankdots: and a `patch`

blankdots: also this might create an edge case with where `/folders/{folderId}` == /folders/publish`

blankdots (changes requested):
some things, will test later

blankdots (approved):

PR #146, opened 2020-11-23
Implement a way to mirror data from EGA

csc-jm-zz:
### Description

This PR aims to implement a solution to mirror data from EGA into the application database with its own `/mirror` endpoint.

### Related issues

Ticket no. 116 from internal board

### Type of change

### Changes Made

- helper methods for requesting data from EGA
- added endpoint for mirroring
- some docstring and typo fixes

### Testing

- [ ] Unit Tests

### Mentions

Script used for mirroring is heavily based on the one [found here](https://github.com/neicnordic/sda-metadata-mirror).

### DIFF
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -890,6 +890,10 @@ class DataMirrorHandler:
         query = MultiDictProxy(MultiDict([("egaStableId", dataset_id)]))
         _, _, _, total_objects = await Operator(db_client).query_metadata_database("dataset", query, 1, 1, [])
         if total_objects > 0:
+            LOG.info(f"Dataset {dataset_id} has already been mirrored into the database.")
+            # Idea with this is to query for a dataset with the exactly same egaStableId value
+            # and if none is found, then it is mirrored from EGA. As of now, this query doesn't work properly
+        else:
             ega_data = MetadataMirror().mirror_dataset(dataset_id)
             operator = Operator(db_client)
             for schema_type in ega_data:
--- a/metadata_backend/api/handlers.py
+++ b/metadata_backend/api/handlers.py
@@ -720,28 +720,23 @@ class RESTApiHandler:
         """
         dataset_id = req.match_info["datasetId"]
         db_client = req.app["db_client"]
-        # Get the dataset from EGA
+        self.data_mirroring(dataset_id, db_client)
+        return web.Response(status=202)
+
+    async def data_mirroring(self, dataset_id: str, db_client: AsyncIOMotorClient) -> None:
+        """Get the dataset and its objects from EGA and POST the metadata objects.
+
+        :param dataset_id: ID of dataset in EGA
+        :param db_client: Motor client used for database connections
+        """
         ega_data = MetadataMirror().mirror_dataset(dataset_id)
-        # POST the metadata objects from the dataset
         operator = Operator(db_client)
-        added = []
         for schema_type in ega_data:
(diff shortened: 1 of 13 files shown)


blankdots:
considering PRs #158 and #151 I think we will need to group at least each dataset related information and make that dataset public so that the model we are working on the DB is followed.

metadata_backend/api/handlers.py
```diff
@@ -461,6 +463,32 @@ async def delete_user(self, req: Request) -> Response:
...
 
+    async def mirror_dataset(self, req: Request) -> Response:
+        """Mirror dataset content queried from EGA into the database.
+
+        :param req: POST request
+        :returns: HTTP No Content response
+        """
+        dataset_id = req.match_info["datasetId"]
+        db_client = req.app["db_client"]
+        await spawn(req, self.data_mirroring(dataset_id, db_client))
```
blankdots, line 474: this returns a Job https://aiojobs.readthedocs.io/en/stable/api.html#job which we can watch to see if it has finished or not.

I think it is perfectly fine if we move these it its own handler class/ file as not sure it reuses anything from the base handler class.

There we could implement also an endpoint where we would check the status of that job (active, pending or closed)

metadata_backend/server.py
```diff
@@ -78,6 +78,7 @@ async def init() -> web.Application:
         web.delete("/users/{userId}", rest_handler.delete_user),
         web.post("/submit", submission_handler.submit),
         web.post("/validate", submission_handler.validate),
+        web.get("/mirror/{datasetId}", rest_handler.mirror_dataset),
```
blankdots, line 81: this endpoint we will need to put behind auth.

metadata_backend/api/handlers.py
```diff
@@ -461,6 +463,32 @@ async def delete_user(self, req: Request) -> Response:
...
+        return web.Response(status=202)
+
+    async def data_mirroring(self, dataset_id: str, db_client: AsyncIOMotorClient) -> None:
+        """Get the dataset and its objects from EGA and POST the metadata objects.
+
+        :param dataset_id: ID of dataset in EGA
+        :param db_client: Motor client used for database connections
+        """
+        ega_data = MetadataMirror().mirror_dataset(dataset_id)
```
blankdots, line 483: we need to check if the database already contains the dataset and if it is in a published folder (because we are mirroring public information, I think the folder should be published by default, but if there is another option fine with that as well) so that we don't mirror it again

metadata_backend/server.py
csc-jm: done

metadata_backend/api/handlers.py
csc-jm: Lines 890-892 of `handlers.py` has an attempt at doing this check. However, the check is not implemented properly yet and needs refining.

metadata_backend/api/handlers.py
csc-jm: The endpoint was moved to its own handler class in `handlers.py`, but job status check endpoint remains unimplemented.

csc-jm:
Unfortunately, I will not be able to work on this PR any further myself so the things that still need to be done here are:
- Fix the check for whether datasets from EGA with a specific `egaStableId` have already been mirrored or not
- Possibly implement the endpoint for checking status of running/finished jobs (this could be merged without it and done in another PR)
- Accessing the endpoint seems to try to save a local file currently which is not desired behavior. The helper methods should only add the mirrored data from EGA to the application database in the background

blankdots:
we will probably not need it for now, closing as we need to mirror data to EGA #295

PR #150, opened 2020-11-27
check idp starts with

### LINKED ISSUE #147: login redirect causes to show forbidden if user logs in after expired session
#### To Reproduce
Steps to reproduce the behaviour:
1. Go to main page
2. Click on login
3. wait for session to expire
4. login again
5. forbidden appears

#### Expected behavior

correct redirect to home page without forbidden

#### Additional context

happens I assume because of lingering cookies or that incorrect checks in the back-end

blankdots:
### Description

idp might contain an redirect and other query string parameters
we need to make sure its base is the same

### Related issues

Closes #147 

### Type of change

### Changes Made

1. Change referer header check to `startswith`
2. fix jsonpatch syntax of path must start with `/`

### Testing

### DIFF
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -49,7 +49,7 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
 
 @middleware
 async def check_login(request: Request, handler: Callable) -> StreamResponse:
-    """Check login if session is user is logged in and can access API."""
+    """Check login if session user is logged in and can access API."""
     if (
         request.path
         in [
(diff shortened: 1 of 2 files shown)


metadata_backend/api/middlewares.py
```diff
@@ -49,7 +49,7 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
 
 @middleware
 async def check_login(request: Request, handler: Callable) -> StreamResponse:
-    """Check login if there is a username."""
+    """Check login if session is user is logged in and can access API."""
```
csc-jm, line 52: Confusing sentence. Maybe this is correct?
```suggestion
    """Check login if session user is logged in and can access API."""
```

csc-jm (approved):
Just one typo possibly. Otherwise seems good.

PR #151, opened 2020-11-27
registered user login access to API and resources

### LINKED ISSUE #135: register user: create or retrieve
#### Description

Once a user is logged in using #133 we need to check if there is a user already in the db with the same parameters if not we create it but if it is we will be using it to create the connection with #106

We will need to figure out how to communicate the registered user to the front-end and while a cookie is an easy method maybe we can find something different.

#### DoD (Definition of Done)

Add function that uses `get_userinfo` and creates user upon callback.

#### Testing

Unit testing and maybe later integration testing.

blankdots:
### Related issues

Fixes #135 

### Type of change

### Changes Made

1. fix oidc workflow jwt_check for IDToken as it was not checking claims
2. create user id if not exists based on eppn (`exists_eppn_user` op add in DB)
3. save user id to session
4. restrict user API to user `current` and retrieve the id from session
5. add mock oidc server
6. modify integration test for login
7. update unit tests
8. fix description, typos
9. in `users.json ` `username` became `name`
10. fixed `controlled_paths` logic to require login
11. restrict login to haka and csc

### Testing

- [x] Unit Tests
- [x] Integration Tests


tests/integration/run_tests.py
```diff
@@ -297,305 +289,306 @@ async def test_patch_drafts_works(schema, filename, filename2):
...
-    await asyncio.gather(*[test_crud_works(schema, file) for schema, file in test_xml_files])
+    async with aiohttp.ClientSession() as sess:
+        LOG.debug("=== Login mock user ===")
+        await login(sess)
 
-    # Test adding and getting draft objects
-    LOG.debug("=== Testing basic CRUD drafts operations ===")
-    await asyncio.gather(*[test_crud_drafts_works(schema, file, file2) for schema, file, file2 in test_json_files])
+        LOG.debug("=== Testing basic CRUD operations ===")
+        # await asyncio.gather(*[test_crud_works(sess, schema, file) for schema, file in test_xml_files])
```
blankdots, line 559: uncomment

csc-jm (approved):

PR #168, opened 2020-12-21
Switch to using UUIDs

### LINKED ISSUE #154: use UUIDs for users, folders and objects IDs
#### Description

Currently users, folders and objects IDs use hex token to identify them, but these don’t guarantee uniqueness
Make use of UUIDs 

#### DoD (Definition of Done)

Generate IDs for users and objects in UUID format

#### Testing

Unit tests

csc-jm:
### Description

PR changes the previous randomly generated hex tokens to UUIDs instead when assigning unique IDs to users, folders and objects.

### Related issues

Fixes #154 

### Type of change

### Changes Made

- changed the id generation methods
- updated the previous unit tests

### Testing

- [x] Unit Tests

### DIFF
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -1,9 +1,9 @@
 """Operators for handling database-related operations."""
 import re
-import uuid
 from abc import ABC, abstractmethod
 from datetime import datetime
 from typing import Any, Dict, List, Tuple, Union
+from uuid import uuid4
 
 from aiohttp import web
 from dateutil.relativedelta import relativedelta
@@ -442,7 +442,7 @@ class Operator(BaseOperator):
 
         Will be replaced later with external id generator.
         """
-        sequence = "".join(str(uuid.uuid4()))
+        sequence = uuid4().hex
         LOG.debug("Generated accession ID.")
         return sequence
 
@@ -774,7 +774,7 @@ class FolderOperator:
 
         :returns: str with folder id
         """
-        sequence = "".join(str(uuid.uuid4()))
+        sequence = uuid4().hex
         LOG.debug("Generated folder ID.")
         return sequence
 
@@ -993,6 +993,6 @@ class UserOperator:
 
         :returns: str with user id
         """
-        sequence = "".join(str(uuid.uuid4()))
+        sequence = uuid4().hex
         LOG.debug("Generated user ID.")
         return sequence
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -1,7 +1,6 @@
 """Operators for handling database-related operations."""
 import re
-import secrets
-import string
+import uuid
 from abc import ABC, abstractmethod
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


metadata_backend/api/operators.py
```diff
@@ -443,9 +442,9 @@ def _generate_accession_id(self) -> str:
 
         Will be replaced later with external id generator.
         """
-        sequence = "".join(secrets.choice(string.digits) for i in range(16))
+        sequence = "".join(str(uuid.uuid4()))
```
csc-jm, line 445: Generating UUID here will result in the standard format that would look for example like this:
`123e4567-e89b-12d3-a456-777888999000`

If we don't want the hyphens, you can get a version if the UUID without the hyphens by replacing the line with this:
`sequence = "".join(uuid.uuid4().hex)`

blankdots:
it is weird in integration tests it wants to connect to `localhost:300`
can we base this on `develop` instead of `feature/data-mirroring` ? that branch has some outstanding issues

metadata_backend/api/operators.py
```diff
@@ -1,7 +1,6 @@
 """Operators for handling database-related operations."""
 import re
-import secrets
-import string
+import uuid
```
blankdots, line 3:
```suggestion
from uuid import uuid4
```

and it will make the syntax simpler

csc-jm:
Oh yea, my bad. Meant to PR this to develop

metadata_backend/api/operators.py
blankdots: even simpler is `uuid4().hex` it generates string by default https://docs.python.org/3/library/uuid.html#uuid.UUID.hex

blankdots: and yes please use `.hex`

blankdots (changes requested):

tests/test_operators.py
```diff
@@ -59,8 +60,8 @@ async def async_patch():
 
         MagicMock.__await__ = lambda x: async_patch().__await__()
         self.client = MagicMock()
-        self.accession_id = "EGA123456"
-        self.folder_id = "FOL12345678"
+        self.accession_id = str(uuid.uuid4())
```
blankdots, line 63:
```suggestion
        self.accession_id = uuid4().hex
```

tests/test_operators.py
```diff
@@ -59,8 +60,8 @@ async def async_patch():
...
+        self.folder_id = str(uuid.uuid4())
```
blankdots, line 64:
```suggestion
        self.folder_id = uuid4().hex
```

tests/test_operators.py
```diff
@@ -2,6 +2,7 @@
 import datetime
 import re
 import unittest
+import uuid
```
blankdots, line 5:
```suggestion
from uuid import uuid4
```

blankdots (changes requested):
keeping tests consistent as well

blankdots (approved):

PR #171, opened 2020-12-23
fix location header for logout redirect

blankdots:
### Description

logout location header was not correct thus this is to change it

### Type of change

### Changes Made

1. changed location header in auth handler logout and in `delete_user`

### Testing

- [x] Tests do not apply


csc-jm (approved):
👍

PR #173, opened 2020-12-29
Feature/add health endpoint

### LINKED ISSUE #156: Add health endpoint
#### Description

Add health endpoint that details the current status of the metadata submitter. the status can be dependent on existing database connection, load times etc.
health endpoint is not restricted to logged in user

#### DoD (Definition of Done)

`/health` endpoint has various status of Ok, Partially down (in case db is down or no connection to dependent services), or Degraded if load times are slow

#### Testing

Unit testing, integration testing and pee review

csc-jm:
### Description

PR adds a health check endpoint.

### Related issues

Fixes #156 

### Type of change

### Changes Made

- Added a health check endpoint
- Added tests for the endpoint

### Testing

- [x] Unit Tests
- [x] Integration Tests

### Mentions

Health check format is based on the one found here in [swift-browser-ui](https://github.com/CSCfi/swift-browser-ui).

### DIFF
--- a/metadata_backend/api/health.py
+++ b/metadata_backend/api/health.py
@@ -8,6 +8,7 @@ from motor.motor_asyncio import AsyncIOMotorClient
 from pymongo.errors import ConnectionFailure
 
 from ..helpers.logger import LOG
+from ..conf.conf import url
 
 
 class HealthHandler:
@@ -19,7 +20,7 @@ class HealthHandler:
         :param req: GET request
         :returns: JSON response containing health statuses
         """
-        db_client = req.app["db_client"]
+        db_client = await self.create_test_db_client()
         services: Dict[str, Dict] = {}
         full_status: Dict[str, Union[Dict, str]] = {}
         services["database"] = {"status": "Ok"} if await self.try_db_connection(db_client) else {"status": "Down"}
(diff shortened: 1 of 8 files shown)


tests/integration/run_tests.py
```diff
@@ -596,15 +606,6 @@ async def main():
...
 
-        # Test add, modify, validate and release action with submissions
-        LOG.debug("=== Testing actions within submissions ===")
-        submission_folder = {
-            "name": "submission test 1",
-            "description": "submission test folder 1",
-        }
-        submission_folder_id = await post_folder(sess, submission_folder)
-        await test_submissions_work(sess, submission_folder_id)
-
```
csc-jm, line 607: This seemed like a duplicate test so removed it... unless there's a reason for it to be there?

blankdots: that was there for a purpose to show that objects separation is done for each user

csc-jm: reverted this commit so it's back in there

metadata_backend/api/health.py
```diff
@@ -0,0 +1,43 @@
...
+class HealthHandler:
+    """Handler for health check."""
+
+    async def get_health_status(self, req: Request) -> Response:
+        """Check health status of the application and return a JSON object portraying the status.
+
+        :param req: GET request
+        :returns: JSON response containing health statuses
+        """
+        db_client = req.app["db_client"]
```
blankdots, line 22: to avoid the 15 retry delay can we create a new AsyncIOMotorClient that uses the `url` from config ?

csc-jm: Attempted this in the latest commits

blankdots (approved):

PR #176, opened 2021-01-19
Hotfix/check exists

blankdots:
### Description

- check objects and folders exist before any operation
- integration check to see if deleted object or folder are still registered in db

### Type of change

### Changes Made

1. added `check_object_exists` to check object exists and fail early with `404` before checking it belongs to user
2. refactor and added more `check_folder_exists` to check folder exists before doing anything 
3. integration test to check objects are deleted properly

### Testing

- [x] Integration Tests

### Mentions

will merge this into develop for now and ask someone to review it one it is released.


blankdots:
pushing this to `develop` and it can be reviewed there, as it is a hotfix

PR #177, opened 2021-01-19
bump to 0.7.1

blankdots:
### Description

hotfix release and bump to `0.7.1`
### Type of change

### Changes Made

see #176 

### Testing

- [x] Unit Tests
- [x] Integration Tests

### DIFF
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -240,12 +240,20 @@ class BaseOperator(ABC):
             raise web.HTTPBadRequest(reason=reason)
         return delete_success
 
-    @abstractmethod
     async def check_exists(self, schema_type: str, accession_id: str) -> None:
         """Check the existance of a object by its id in the database.
 
-        Must be implemented by subclass.
+        :param schema_type: Schema type of the object to find.
+        :param accession_id: Identifier of object to find.
+        :raises: HTTPNotFound if object does not exist
+        :returns: None
         """
+        exists = await self.db_service.exists(schema_type, accession_id)
+        LOG.info(f"check_exists: {exists}")
+        if not exists:
+            reason = f"Object with id {accession_id} from schema {schema_type} was not found."
+            LOG.error(reason)
+            raise web.HTTPNotFound(reason=reason)
 
     @abstractmethod
     async def _format_data_to_create_and_add_to_db(self, schema_type: str, data: Any) -> str:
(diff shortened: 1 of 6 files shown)


tests/test_operators.py
```diff
@@ -775,12 +775,12 @@ async def test_folder_object_remove_fails(self):
         with self.assertRaises(HTTPBadRequest):
             await operator.remove_object(self.test_folder, "study", self.accession_id)
 
-    async def test_check_folder_exists_fails(self):
+    async def testcheck_folder_exists_fails(self):
```
teemukataja, line 778: Does this need an underscore `_` to `testcheck` -> `test_check` to be discovered by the test suite?

metadata_backend/api/operators.py
```diff
@@ -488,6 +495,21 @@ def format_date(key: str, doc: Dict) -> Dict:
             doc = format_date("publishDate", doc)
         return doc
 
+    async def check_exists(self, schema_type: str, accession_id: str) -> None:
```
teemukataja, line 498: `Operator(BaseOperator)` and `XMLOperator(BaseOperator)` both use `check_exists()` and they both inherit `BaseOperator`, so can `check_exists()` be moved to `BaseOperator` class, so that the function is not duplicated?

metadata_backend/api/operators.py
```diff
@@ -567,6 +589,20 @@ async def _format_read_data(self, schema_type: str, data_raw: Dict) -> str:
         """
         return data_raw["content"]
 
+    async def check_exists(self, schema_type: str, accession_id: str) -> None:
```
teemukataja, line 592: `Operator(BaseOperator)` and `XMLOperator(BaseOperator)` both use `check_exists()` and they both inherit `BaseOperator`, so can `check_exists()` be moved to `BaseOperator` class, so that the function is not duplicated?

teemukataja (approved):
Approved, but see questions

metadata_backend/api/operators.py
blankdots: it is an a `ABC` https://docs.python.org/3/library/abc.html and the method needs to be implemented in the subclass

metadata_backend/api/operators.py
blankdots: it is an a `ABC` https://docs.python.org/3/library/abc.html and the method needs to be implemented in the subclass

tests/test_operators.py
blankdots: will work without it, but reverting to keep naming convention.

metadata_backend/api/operators.py
blankdots: but it is a good point I can reimplement it a bit better as it is used in other places as well like user and `folder`.

blankdots: previously when i had it as a reusable function it complained now it does not :shrug:

Issue #178, opened 2021-01-28
Support serving SSL from metadata-submitter


pontus:
#### Proposed solution

It would be useful to be able to serve https from the metadata-submitter directly. There is the obvious use case (which I consider bad) of running the service without any proxy in front. Our use case is to avoid having clear text traffic, also in the internal network inside our kubernetes setup.

It seems gunicorn can do https termination nowadays, so it should be possible to check for variables (e.g. pointing to certificate, key and CA) and add corresponding options to pass into gunicorn. 

#### DoD (Definition of Done)

gunicorn in the prebuilt container can be started with TLS-parameters and https can be used to communicate with it.

#### Testing

Use https to connect to the port for the service.

PR #180, opened 2021-01-29
Support TLS from container

pontus:
### Description

Provides a solution for #178 by looking for specific environment variables and passing corresponding options to gunicorn.

### Related issues

#178 

### Type of change

### Changes Made

1. Documentation updates.
2. Changes to app launch script to pass additional options to gunicorn.
3. Github action to test production container in basic ways (with and without TLS).
 
### Testing

- [ ] Unit Tests
- [x] Integration Tests
- [ ] Tests do not apply


blankdots:
This is a good option, the remaining 2 options is that we can server HTTPS from the aiohttp, but the genera recommendation would be as pointed out in the issue to server a proxy in front of the service and i think that should me integrated in the helm chart.

blankdots (approved):

PR #185, opened 2021-02-02
if eppn is missing resolve sub claim

blankdots:
### Description

the `sub` claim is used by Elixir AAI userinfor endpoint

### Type of change

### Changes Made

1. check for `sub` claim if `eppn` is missing, otherwise raise exception

### Testing


pontus (approved):
Looks solid. Service also gets farther along.

PR #203, opened 2021-02-15
Implement support for multiple simultaneous sessions

pontus:
### Description

Add support for multiusers (multiple simultaneous sessions).

One could consider doing this with a library to provide sessions, but as a non-core author, I aim to keep this PR as minimal as possible.

While I believe these changes to be correct, I do note that there it touches a lot of lines in important areas, so a bit of a warning that it might be recommend to take some time for review. 

### Type of change

### Changes Made

* Make sessions references indirect (accessible via session id,taken from cookie).
* Change

### Testing

- [x] Unit Tests
- [ ] Integration Tests
- [ ] Tests do not apply

### DIFF
--- a/metadata_backend/api/auth.py
+++ b/metadata_backend/api/auth.py
@@ -7,14 +7,14 @@ import json
 
 from aiohttp import web, BasicAuth, ClientSession
 from aiohttp.web import Request, Response
-from .middlewares import decrypt_cookie, generate_cookie, get_session
+from .middlewares import decrypt_cookie, generate_cookie
 from authlib.jose import jwt
 from authlib.oidc.core import CodeIDToken
 from authlib.jose.errors import MissingClaimError, InvalidClaimError, ExpiredTokenError, InvalidTokenError, DecodeError
 from multidict import CIMultiDict
 from .operators import UserOperator
 
-from typing import Dict, Tuple, Union
+from typing import Dict, Tuple
 
 from ..helpers.logger import LOG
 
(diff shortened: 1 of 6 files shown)


blankdots:
a good library would be https://github.com/aio-libs/aiohttp-session/ which we considered, however similar functionality could be achieved without using that library (which i think is implemented with the current version), well there is a need to go redis.

blankdots:
with this AccessHandler methods `_save_to_session` and `_get_session` are obsolete do i understand it correctly ?

pontus:
Correct, those are obsoleted (but were left as I thought they were still in use). Removed by 45a0112.

metadata_backend/api/auth.py
```diff
@@ -87,9 +88,8 @@ async def callback(self, req: Request) -> Response:
             LOG.error(reason)
             raise web.HTTPBadRequest(reason=reason)
 
-        # Verify, that states match
-        state = await self._get_from_session(req, "oidc_state")
-        if not secrets.compare_digest(str(state), str(params["state"])):
+        # Verify, that state is pending
+        if not params["state"] in req.app["OIDC_State"]:
```
blankdots, line 92: wondering why we don't need to compare digest here or why we dropped it ?

metadata_backend/api/middlewares.py
```diff
@@ -30,6 +30,7 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
     except web.HTTPError as error:
         details = _json_exception(error.status, error, req.url)
         LOG.error(details)
+        # LOG.error(error.__traceback__.format_exc())
```
blankdots, line 33: forgotten log ?

metadata_backend/api/auth.py
pontus: With this PR, we delay session set up (cookie creation) to the callback, so we don't have the connection to compare the specific handed out `state` to the received. As a second best thing, we keep track of currently outstanding authentications (where callbacks have not yet been called).

I'm not sure I see understand what the security benefit of verifying the state is, but if one want to keep track of it, one can move the session creation (cookie setting) to `login`. The drawback of that (reason I did not do it for this PR) is that the cookie can not be completely filled in at that time and I've seen overall weird behaviour when changing cookies (cookies on redirects are also a thing, and I don't have a complete overview for what cases it works and when it's ignored).

metadata_backend/api/middlewares.py
pontus: Yes, apologies.

blankdots (approved):
I think we can have this merged into develop. No objections.

PR #231, opened 2021-06-16
Folder, draft and draft template pagination

### LINKED ISSUE #222: Pagination for folders, drafts and user drafts (templates)
#### Description

We would like that the response `/folders` would be paginated so it is easier to display it in the UI
The same goes for `/user/{userid}` drafts (templates) - for this possibly require a new endpoint  `/user/{userid}/drafts`

#### DoD (Definition of Done)

Pagination is implemented for for folders, drafts and user drafts (templates)

#### Testing

Unit and Integration tests

csc-jm:
### Description

PR adds `page` key with proper pagination values into the JSON object returned by `/folders`. The values can be changed with URL query parameters eg. `/folders?page=2&per_page=10&published=true`. Default `per_page` value for folders is currently 5. Also a paginated list of user drafts and the folder list from the user object can be now requested with eg. `/users/current?items=drafts`.

### Related issues

Fixes #222 

### Type of change

- [x] New feature (non-breaking change which adds functionality) **(however adjustments in frontend are required)**

### Changes Made

- changed `get_folders` method to do as described above
- changed `update_folder` method to take pagination into account when querying
- changed `get_user` method to return only drafts or folder IDs if queried specifically
- made `_header_links` method shared
- made `_get_page_param` method shared
- changed previous tests and added new ones
- updated openapi specs

### Testing

- [x] Unit Tests
- [x] Integration Tests


metadata_backend/api/handlers.py
```diff
@@ -1062,3 +1089,24 @@ async def _extract_xml_upload(req: Request, extract_one: bool = False) -> List[T
             files.append((xml_content, schema_type))
             LOG.debug(f"processed file in {schema_type}")
     return sorted(files, key=lambda x: schema_types[x[1]]["priority"])
+
+
+def _header_links(url: str, page: int, size: int, total_objects: int) -> CIMultiDict[str]:
```
blankdots, line 1094: any reason not to have this as part of `RestAPIHandler` class?

blankdots (commented):
did you manage to test this frontend to see if there are any breaking changes there?

metadata_backend/api/handlers.py
```diff
@@ -523,24 +503,71 @@ def _check_patch_folder(self, patch_ops: Any) -> None:
...
         current_user = get_session(req)["user_info"]
         user = await user_operator.read_user(current_user)
 
-        operator = FolderOperator(db_client)
-        folders = await operator.query_folders({"folderId": {"$in": user["folders"]}})
+        folder_query = {"folderId": {"$in": user["folders"]}}
+        # Check if only published or draft folders are requestsed
+        if "published" in req.query:
+            pub_param = req.query.get("published", "").lower()
+            if pub_param == "true":
```
blankdots, line 538: we can check pub_param in list of true or false then in the query we convert pub_param to boolean to simplify this check

metadata_backend/api/handlers.py
csc-jm: Good point

metadata_backend/api/handlers.py
csc-jm: Good point. So like this?
```suggestion
            if "published" in req.query:
                pub_param = req.query.get("published", "").title()
                if pub_param in ["True", "False"]:
                    folder_query["published"] = {"$eq": eval(pub_param)}
```

blankdots: `pub_param in ["True", "False"]:` maybe lowercase/standardise the value as we don't know what the input will look like

csc-jm: `.title()` capitalizes first letter of a string so the idea being that inputs like `"false"`, `"FALSE"` or `"faLse"` would all become `"False"`

blankdots: i see - well ok

tests/integration/run_tests.py
```diff
@@ -1081,6 +1160,16 @@ async def main():
...
+    # Remove the remaining user in the test database
+    async with aiohttp.ClientSession() as sess:
+        await login(sess, other_test_user, other_test_user_given, other_test_user_family)
+        async with sess.get(f"{users_url}/{user_id}") as resp:
+            LOG.debug(f"Reading user {user_id}")
+            assert resp.status == 200, "HTTP Status code error"
+            response = await resp.json()
+            real_user_id = response["userId"]
+        await delete_user(sess, real_user_id)
+
```
csc-jm, line 1172: Adding this allowed me to rerun integration tests multiple times but is not a good long-term solution for dropping database/collection after integration tests have completed.

metadata_backend/server.py
```diff
@@ -92,6 +92,7 @@ async def init() -> web.Application:
         web.get("/users/{userId}", _user.get_user),
```
blankdots, line 92: just a wild idea, how about we add query string parameters to this endpoint so that we can do something like:
`GET /users/current?type=drafts` i put `type` for lack of inspiration but the idea is to list both `folders` and `drafts` even though for folders it will be just a list of strings

csc-jm: I have named this parameter `items` now, so `/users/current?items=drafts` returns the paginated list of drafts from the user object and `/users/current?items=folders` returns the list of folder IDs from the user object.

tests/integration/run_tests.py
blankdots: i think for completing this issue will have to do let us create a new issue to clean integration tests.

blankdots: https://github.com/CSCfi/metadata-submitter/issues/234

tests/integration/run_tests.py
```diff
@@ -1081,6 +1200,16 @@ async def main():
...
 
+    # Remove the remaining user in the test database
+    async with aiohttp.ClientSession() as sess:
+        await login(sess, other_test_user, other_test_user_given, other_test_user_family)
+        async with sess.get(f"{users_url}/{user_id}") as resp:
+            LOG.debug(f"Reading user {user_id}")
+            assert resp.status == 200, "HTTP Status code error"
+            response = await resp.json()
+            real_user_id = response["userId"]
+        await delete_user(sess, real_user_id)
```
csc-jm, line 1211: Necessity of this is questionable

blankdots: either is fine for me, but if we are not using or we don't need some code, good practice is to remove it.

blankdots (approved):

PR #235, opened 2021-07-01
Load tests

### LINKED ISSUE #79: Profile query implementation with large collections
#### Description
We don't currently now how well queries do perform for large collections, so some performance testing need to be done at some point.

csc-jm:
### Description

After `pip install locust`, these tests can be run on the project root by:
```
locust --tags post  # this executes tests tagged with "post" which post folders and objects
locust --tags query  # this executes tests that query for folders and objects
```

The values set in the `locust.conf` file can be overridden by adding the same tags to the end of the terminal command [as listed here](https://docs.locust.io/en/stable/configuration.html#all-available-configuration-options).

### Related issues

Fixes #79 

### Type of change

### Testing

- [x] Tests do not apply


tests/locustfile.py
```diff
@@ -0,0 +1,128 @@
+"""Run load tests for submitting folders and objects and querying them."""
```
blankdots, line 1: i think we can create a special folder just for these files called `performance` under `tests`

blankdots (approved):
I think we should add documentation under testing https://metadata-submitter.readthedocs.io/en/develop/test.html how to run these tests and this issue can be considered done.

Tests represent a good start and give us plenty info.

Approving given that the documentation will be done.

PR #236, opened 2021-07-16
Enable CSV uploading

### LINKED ISSUE #221: Upload objects as CSV
#### Description

Currently we support uploading data as XML and JSON, however we would like to enable data upload as CSV

Most likely we would like to support CSV for objects where we know data will this will be necessary such as `Run`, `Analysis` and  others.

#### DoD (Definition of Done)

API supports CSV upload

#### Testing

Unit and Integration testing

csc-jm:
### Description

This is the first part of working with CSV files.
In practice we will support a limited number of objects for CSV, and the first one is `sample` according to EGA schema.

### Related issues

Fixes #221 

### Type of change

### Changes Made

1. allow submitting `sample` objects as CSV - we support multiple sample objects all in the same row.
2. introduce CSV parser that can work with `sample` object, CSV file content
3. add integration and unit tests for CSV
4. `server.py` - add state configuration on startup event  https://docs.aiohttp.org/en/stable/web_advanced.html#background-tasks
5. clean-up and small fixes

### Testing

- [x] Unit Tests
- [x] Integration Tests

### Mention
follow-up issues:
- #321 
- #322


metadata_backend/api/handlers.py
```diff
@@ -1107,19 +1108,19 @@ def setup_static(self) -> Path:
...
 
     :param req: POST request containing "multipart/form-data" upload
-    :raises: HTTPBadRequest if request is not valid for multipart or multiple files sent. HTTPNotFound if
-    schema was not found.
-    :returns: content and schema type for each uploaded file, sorted by schema
-    type.
+    :param extract_one: boolean stating whether multiple files should be handled
+    :param is_xml: boolean stating if file can be expected to be XML
+    :raises: HTTPBadRequest for multiple different reasons
+    :returns: Either list of xml files or csv files for further parsing
```
blankdots, line 1120: to my understanding it returns content not a list of files, did i understand it correctly?

metadata_backend/api/handlers.py
```diff
@@ -1133,23 +1134,72 @@ async def _extract_xml_upload(req: Request, extract_one: bool = False) -> List[T
...
-                    break
-                data.append(chunk)
-            xml_content = "".join(x.decode("UTF-8") for x in data)
-            files.append((xml_content, schema_type))
-            LOG.debug(f"processed file in {schema_type}")
-    return sorted(files, key=lambda x: schema_types[x[1]]["priority"])
+
+        content, schema_type = await _extract_upload(part)
+
+        if is_xml:
```
blankdots, line 1144: i think we should split this into its own function, or at least split `_multipart_content` into checking file type and returning file/content

metadata_backend/api/handlers.py
```diff
@@ -1133,23 +1134,72 @@ async def _extract_xml_upload(req: Request, extract_one: bool = False) -> List[T
...
+            # Validation is skipped here because validation endpoint
+            # does its own validating on a supposed XML
+            xml_files.append((content, schema_type))
+        else:
+            # Multipart can have XML or CSV files
+            # Default expectation is XML, so XML is tried first
+            schema = XMLSchemaLoader().get_schema(schema_type)
```
blankdots, line 1151: to test if it is XML you migth be able to use: ` XMLResource` from `xmlschema` package as you can see in tests: https://github.com/sissaschool/xmlschema/blob/bb528a94d83f9fc937765d16e266506bcf127318/tests/test_resources.py#L492-L494 `ParseError` is triggered when xml is not valid.  That will avoid us needing to load a schema.

Also a `try` `except` might be better here where we catch exact error e.g. if we ` except csv.Error:` then we do in that except a `try` with xml if that gives `except xmlschema.etree.ParseError:` then we fail as unknown format

tests/integration/run_tests.py
```diff
@@ -1109,8 +1141,8 @@ async def main():
             "description": "submission test folder 1",
         }
         submission_folder_id = await post_folder(sess, submission_folder)
-        await test_get_folders(sess, submission_folder_id)
-        await test_get_folders_objects(sess, submission_folder_id)
+        # await test_get_folders(sess, submission_folder_id)
```
blankdots, line 1144: I assume this is still in progress, or any other reason these lines were commented ?

tests/test_files/sample/EGAformat.csv
```diff
@@ -0,0 +1,2 @@
+title,alias,description,subjectId,bioSampleId,caseOrControl,gender,organismPart,cellLine,region,phenotype
+test sample,test1,A test sample metadata CSV file based on the EGA submitter portal format,123456789abc,,,unknown,,,,some disease state
```
blankdots, line 2: we need more lines here, as in real world scientists would have multiple lines for this.

blankdots (changes requested):
i think `_multipart_content` needs some rework.

would be nice to have more csv unit tests not only for sample but also for the file formats.

metadata_backend/api/handlers.py
```diff
@@ -316,9 +315,11 @@ async def post_object(self, req: Request) -> Response:
         content: Union[Dict, str]
         operator: Union[Operator, XMLOperator]
         if req.content_type == "multipart/form-data":
-            files = await _extract_xml_upload(req, extract_one=True)
+            files, type = await _multipart_content(req, extract_one=True)
```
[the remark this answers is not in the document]
csc-jm: Good point. Changed to `cont_type`

metadata_backend/api/handlers.py
```diff
@@ -1133,23 +1134,72 @@ async def _extract_xml_upload(req: Request, extract_one: bool = False) -> List[T
...
+            LOG.info(f"{schema_type} schema loaded.")
+            validator = XMLValidator(schema, content)
+            if validator.is_valid:
+                LOG.info("Valid XML content was extracted.")
+                xml_files.append((content, schema_type))
+            # TODO this is still not working !!
+            else:
+                LOG.info("Request does not contain valid XML. Trying CSV extracting instead.")
+                parser = CSVToJSONParser()
+                if parser.is_csv(content):
```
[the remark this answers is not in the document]
csc-jm: Good point. Removed the method and included the test in `handlers.py`

metadata_backend/api/handlers.py
csc-jm: Yes. Technically it could be list of contents if multiple XML files are extracted at once but these comments were outdated from a previous commit and forgot to edit.

tests/integration/run_tests.py
csc-jm: Whoops. Used this a bypass before the existence of `clean_db.py` and forgot to undo the commenting out.

metadata_backend/api/handlers.py
csc-jm: Good point. I have separated the content type check to another helper function

metadata_backend/api/handlers.py
```diff
@@ -1133,23 +1134,72 @@ async def _extract_xml_upload(req: Request, extract_one: bool = False) -> List[T
...
-    return sorted(files, key=lambda x: schema_types[x[1]]["priority"])
+
+        content, schema_type = await _extract_upload(part)
+
+        if is_xml:
+            # Validation is skipped here because validation endpoint
+            # does its own validating on a supposed XML
+            xml_files.append((content, schema_type))
+        else:
+            # Multipart can have XML or CSV files
```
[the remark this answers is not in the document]
csc-jm: True, but found issues with this.

metadata_backend/api/handlers.py
csc-jm: Trying `XMLResource` seems to work well. I changed this to try/excepts but as for the order of testing, the key issue is that validating CSV is difficult when something like `<SAMPLE_SET>,</SAMPLE_SET>` is instantly identifiable to us humans as XML but also clearly passes as valid CSV. Hence, with the current way, XML would need to be validated first and CSV after that.

tests/test_files/sample/EGAformat.csv
csc-jm: Could you elaborate bit more? The way CSV is parsed atm is that each line of CSV creates one object so multiple lines would indicate multiple different samples in this case. I understand the need, but I'm wondering about implementation.

blankdots: yes, we our users will upload multiple samples in a csv, thus we need to be able to work with that.

metadata_backend/api/handlers.py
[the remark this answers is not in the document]
blankdots: ok

metadata_backend/api/handlers.py
```diff
@@ -1133,23 +1137,87 @@ async def _extract_xml_upload(req: Request, extract_one: bool = False) -> List[T
...
+            # Multipart request can have XML or CSV files
+            file_type = _check_file_type(content)
+            if file_type == "csv":
+                # Parse CSV to JSON
+                csv_content = CSVToJSONParser().parse(schema_type, content)
+                csv_files.append((csv_content, schema_type))
+            elif file_type == "xml":
+                xml_files.append((content, schema_type))
+
+    if xml_files and csv_files:
```
blankdots, line 1160: lines 1160 - 1173 might deserver their own function as well.

genie9 (approved):

PR #301, opened 2021-11-29
Feature/update dev build

genie9:
### Description

This commit adds the use of .env file for `docker-compose up` so that developer's own environment variables would not be accidentally pushed to the repository. Also makes it easy to handle your own development environment.

### Type of change

### Changes Made
Docker-compose and docker-compose-tls files changed to use variables from .env file. 
Adds .env.example file to copy it to user's local .env file. 
Updates README with instructions.
Updates .gitignore to ignore .env file
Updates github workflow commands on docker-compose with env file flag

Additional changes:
Update README with instructions on Python requirements.
Update pre-commit script to execute tests even if .tox folder does not exist. 

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply

### DIFF
--- a/.github/workflows/int.yml
+++ b/.github/workflows/int.yml
@@ -26,7 +26,7 @@ jobs:
         
     - name: Start Services
       run: |
-        docker-compose up -d --build
+        docker-compose --env-file .env.example up -d --build 
         sleep 30
   
     - name: Run Integration test
@@ -55,7 +55,7 @@ jobs:
 
     - name: Start services with TLS enabled
       run: |
-        docker-compose -f docker-compose-tls.yml up -d
+        docker-compose -f docker-compose-tls.yml --env-file .env.example up -d
         sleep 30
 
     - name: Run Integration test
--- a/.env.example
+++ b/.env.example
@@ -21,6 +21,10 @@
       MONGO_AUTHDB=admin
       MONGO_INITDB_ROOT_PASSWORD=admin
       MONGO_INITDB_ROOT_USERNAME=admin
+      MONGO_SSL=true
+      MONGO_SSL_CA=/tls/cacert
+      MONGO_SSL_CLIENT_KEY=/tls/key
+      MONGO_SSL_CLIENT_CERT=/tls/cert
 
       # doi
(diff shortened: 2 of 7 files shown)


blankdots (commented):
Looks good, I think we might want to do the same kind of treatment to `docker-compose-tls.yml` to support `.env` file

For github actions we might need to add:
- ` --env-file .env.example` to  https://github.com/CSCfi/metadata-submitter/blob/develop/.github/workflows/int.yml#L29 and https://github.com/CSCfi/metadata-submitter/blob/develop/.github/workflows/int.yml#L58

blankdots:
this might also imply changes in the https://github.com/CSCfi/metadata-submitter-frontend/blob/master/.github/workflows/e2etests.yml#L38 line so before we merge this we need to update that as well

.github/workflows/int.yml
```diff
@@ -26,8 +26,8 @@ jobs:
         
     - name: Start Services
       run: |
-        docker-compose up -d --build
-        sleep 30
+        docker-compose up -d --build --env-file .env.example
```
blankdots, line 29:
```suggestion
        docker-compose --env-file .env.example up -d --build
```
the file s envhould come first as per docker-compose syntax

.github/workflows/int.yml
```diff
@@ -55,8 +55,8 @@ jobs:
 
     - name: Start services with TLS enabled
       run: |
-        docker-compose -f docker-compose-tls.yml up -d
-        sleep 30
+        docker-compose -f docker-compose-tls.yml up -d --env-file .env.example
```
blankdots, line 58:
```suggestion
        docker-compose -f docker-compose-tls.yml --env-file .env.example up -d --build
```

blankdots (commented):
also `docker-compose-tls,yml` file should have similar environment variables as `docker-compose.yml`

.github/workflows/int.yml
genie9: thanks!

blankdots (approved):

Issue #280, opened 2021-11-04
enable search by name and date in folders


blankdots:
#### Description

As a user that has many folders, drafts and template i would need to filter them so i can find what i am looking for.

`dateCreated` key already exists in folder but maybe this might need some indexes created in mongodb so that folders can be searched by name (text search)

Related to issue CSCfi/metadata-submitter-frontend/issues/423

#### DoD (Definition of Done)

user is able to filter by name and by date created

#### Testing

unit and integration testing

blankdots:
for dateCreated users might want to search for:
- year
- year and moth
- year, month and day

genie9:
Accidentally conversation held in https://github.com/CSCfi/metadata-submitter-frontend/issues/423#issuecomment-987906669:

@genie9 i assume we are talking about this backend issue CSCfi/metadata-submitter/issues/280

    @blankdots Some questions have arisen during the implementation of folder filtering by name:

    1. Should there be an additional field for searching `text_name` separating folder name by delimiters?
       
       * discussed in RC => yes

true, but that is an implementation detail.

    2. Is it good enough to search by full string matching with text search (indexed)?

yes, text search should be it to take advantage of indexing, we will not do regex. Simple (text) search, we will do, no need to do anything fancy like elastic search does.

    3. Should we implement regex with ^-prefix-  so we could use partial and also indexed search?
       
       * discussed in RC => 2 is chosen

No regex please, it is not required for this use case.

    4. Would name filtering be using only one word, not an array of words?

based on this, if there are multiple words https://docs.mongodb.com/manual/text-search/#-text-operator will create an OR so we can get as many words as needed

    5. Should `description`-field be included in the search (indexed)? How long are descriptions usually and would it mess filtering by name? (Asking because Stefan's example included `description`-field in indexing)

This I am not 100% sure, but I would leave out, as descriptions can often take the form of an abstract. If we get request to do it then we will add it with another issue.

    6. Should the new field `text_name` be shown in the query results?

No. that is internal

PR #305, opened 2021-12-09
Feature/find by name

genie9:
### Description
Adds functionality to filter folders by folder name eg. `/folders?name="test string"`.
The folder collection is updated with the new field `text_name` which is populated during folder creation from folder name as delimiter separated words e.g 'test_file' becomes 'test file'. This field is indexed so a text search would be possible.
The database is indexed during database container initialization.

### Related issues
Feature #280 
- Filter folders by name

### Type of change

### Changes Made
Add indexing on database initialization
- Add docker-compose configuration for mongo container to run a script on
newly initialized database.
- Add js script to build database collections and indexes.

Add new field `text_name` to folder collection
- Folder is updated with the new field `text_name` which is populated
during folder creation from folder name as delimiter separated words.
- Schema is updated with a new field.

Add folder querying by name.

Python scripts for database operations
- Add script mongo_indexes.py for collections and indexes creation to be run if
the database is destroyed.
- Update clean_db.py script with new functionality to only delete
documents from collections. Old functionality ergo dropping database
can be run with the flag `--purge`.

Add openapi specs.

Minor changes:
- Update dev requirements and add the use of env vars in scripts and integration test
to make it easier to run stuff in the local environment.
- Update Github workflow file with env vars.
- Add black formatting argument

### Testing
- [x] Integration Tests

### DIFF
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -692,7 +692,7 @@ class FolderOperator:
             {"$sort": {"dateCreated": -1}},
             {"$skip": skips},
             {"$limit": page_size},
-            {"$project": {"_id": 0}},
+            {"$project": {"_id": 0, "text_name": 0}},
         ]
         data_raw = await self.db_service.do_aggregate("folder", _query)
 
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -3,7 +3,7 @@ import re
 import time
 from abc import ABC, abstractmethod
 from datetime import datetime
-from typing import Any, Dict, List, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 from uuid import uuid4
 
 from aiohttp import web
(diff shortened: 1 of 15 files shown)


tests/integration/run_tests.py
```diff
@@ -925,6 +926,41 @@ async def test_getting_paginated_folders(sess):
...
+    """
+    names = [" filter new ", "_filter_", "-filter-", "_extra-", "_2021special_"]
+    folders = []
+    for name in names:
+        folder_data = {"name": f"Test{name}name", "description": "Test filtering name"}
+        folders.append(await post_folder(sess, folder_data))
+
+    async with sess.get(f"{folders_url}?name=filter") as resp:
+        ans = await resp.json()
+        assert resp.status == 200, f"HTTP Status code error {resp.status} {ans}"
```
blankdots, line 942: i think we need to assert status 200 before the reading the response, and i see this is used differently below, sometimes we check status first then the response, sometimes the other way around

blankdots (approved):
small suggestions of improvement and maybe we want to update the docs with this PR or a follow-up one, namely the https://github.com/CSCfi/metadata-submitter/blob/master/docs/test.rst#integration-testing section

tests/integration/run_tests.py
genie9: this is done because tests were failing and needed to get an error message. Without returning the response message there was no real explanation for what went wrong. This actually works even it's not consistent with other tests.  I thought to leave it like that and maybe continue doing so?

blankdots: then good, let us leave it like that.

metadata_backend/api/operators.py
```diff
@@ -691,7 +692,7 @@ async def query_folders(self, query: Dict, page_num: int, page_size: int) -> Tup
             {"$sort": {"dateCreated": -1}},
```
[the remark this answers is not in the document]
genie9: I wonder how to act on this. This sounds reasonable but I wonder what is the user's anticipation on search: should the date be the first parameter... For example folders are:
"Test folder users" - created 2020 
"Test folder" - created 2019

If searched with `test folder` and sorted by date and then score results are:
"Test folder users" - created 2020 
"Test folder" - created 2019
which feels more correct.

And sort by score then date would be:
"Test folder" - created 2019
"Test folder users" - created 2020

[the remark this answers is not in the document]
blankdots: i am not sure about user's expectations, but I think the following might address any use case we might have:

1. no `name` or `date` filter added - we short by ` {"$sort": {"dateCreated": -1}}`
2. `name` paramter added, as you gave the example we sort by date 
>"Test folder users" - created 2020
> "Test folder" - created 2019
>which feels more correct.
3. `name` and `parameter` added we use both to sort

i think we can combine sort e.g. `{ $sort: { score: { $meta: "textScore" }, dateCreated: -1 } }`

https://docs.mongodb.com/manual/reference/operator/aggregation/sort/#text-score-metadata-sort

blankdots (approved):
looks very good.

PR #315, opened 2021-12-22
Refactor/auth oidcrp

### LINKED ISSUE #157: Explore integration with JWTConnect-Python-OidcRP
#### Description

Currently we are using custom code for OIDC authentication and validation of id token, and with this issue we would like to explore https://github.com/IdentityPython/JWTConnect-Python-OidcRP to see if it brings any advantages or makes it easier to maintain

#### DoD (Definition of Done)

Decision on where to use module, if it is positive then implementation is required

#### Testing

Peer review, if it is positive then unit testing and integration tests

teemukataja:
### Description
Refactor `auth.py` package by removing custom OIDC code and replacing it with [oidcrp](https://github.com/IdentityPython/JWTConnect-Python-OidcRP).

### Related issues
Fixes #157

### Changes Made
- Deprecated ENVs `ISS_URL`, `AUTH_URL`, `AUTH_REFERER`, `JWK_URL`
- New mandatory ENV `OIDC_URL`
- New optional ENVs `OIDC_SCOPE`, `AUTH_METHOD`
- Removed `Authlib` dependency
- Added `oidcrp` dependency

### Testing

- [x] Manual Testing
- [x] Unit Tests
- [x] Integration Tests


metadata_backend/api/auth.py
```diff
@@ -146,11 +147,24 @@ async def callback(self, req: Request) -> Response:
...
         session_id = cookie["id"]
 
-        req.app["Session"][session_id] = {"oidc_state": params["state"], "access_token": access_token}
+        req.app["Session"][session_id] = {"oidc_state": params["state"], "access_token": session["token"]}
         req.app["Cookies"].add(session_id)
-        req.app["OIDC_State"].remove(params["state"])
 
-        await self._set_user(req, session_id, access_token)
+        user_data: Tuple[str, str]
+        if "eppn" in session["userinfo"]:
```
blankdots, line 154: for SDS-AAI this changed a tad see: https://github.com/CSCfi/metadata-submitter/commit/47ba615bec6a078ed18053635b0a451afa7e8643#diff-ebe74c356287148a91f17f859c1f7f9cac90af19df48b372bc1a83d30af76c06R201-R204 `CSCUserName` and `remoteUserIdentifier` might be needed now

blankdots (commented):
looks good so far, but would need the integration tests to tell, if it works as intedend.
maybe we could use this in integration tests: https://github.com/IdentityPython/oidc-op ?

tests/integration/run_tests.py
```diff
@@ -52,7 +52,8 @@
     ("analysis", "ERZ266973.json", "ERZ266973.json"),
 ]
 base_url = os.getenv("BASE_URL", "http://localhost:5430")
-mock_auth_url = os.getenv("OIDC_URL", "http://localhost:8000")
+# mock_auth_url = os.getenv("OIDC_URL", "http://localhost:8000")
```
blankdots, line 55: that means this OIDC_URL is set somewhere ? - why do we need to comment it out here?
I see in the github actions is set to localhost:8000 and i assume locally will work with default value.

tests/integration/mock_auth.py
```diff
@@ -114,28 +124,117 @@ async def jwk_response(request: web.Request) -> web.Response:
...
     return web.json_response(user_info)
 
 
+async def oidc_config(request: web.Request) -> web.Response:
+    """Return standard OIDC configuration."""
+    oidc_config_json = {
+        "issuer": "http://mockauth:8000",
+        # must be localhost to be accessible outside of docker-network
+        "authorization_endpoint": "http://localhost:8000/authorize",
+        "token_endpoint": "http://mockauth:8000/token",
```
blankdots, line 154: `http://mockauth:8000` maybe it can be a parameter ? the same for `localhost:8000`

Dockerfile-dev
```diff
@@ -13,6 +13,7 @@ COPY requirements.txt .
 COPY metadata_backend/ ./metadata_backend
 
 RUN pip install .
+RUN pip install authlib==0.15.5  # required for mockauth (integration test)
```
blankdots, line 16: i would not fix the version, as it would be difficult to keep up to date

blankdots (approved):
some small suggestions of improvement.

Dockerfile-dev
teemukataja: removed version

tests/integration/mock_auth.py
teemukataja: parameterised URLs

tests/integration/run_tests.py
teemukataja: There was a problem, where the integration test would not run, because it was started from outside, and `OIDC_URL` was set to `mockauth` which was only resolvable from inside the docker-network. I differentiated these now to two different URLs, which can also be used in the OIDC config
https://github.com/CSCfi/metadata-submitter/pull/315/commits/8d81026c4e5704653aff390da3abba8050781225#diff-9049ef8d86388471b49a45e70f5427febf75a8d5fb0873132422e2f132b89d34R47-R48

genie9 (commented):
https://github.com/CSCfi/metadata-submitter/pull/315#discussion_r782076219

Dockerfile-dev
teemukataja: I initially tried having authlib in `requirements-dev`, but it didn't help. I found out, `authlib` must be installed in the docker image before for this line to run https://github.com/CSCfi/metadata-submitter/blob/develop/docker-compose.yml#L63

I'll remove the version

genie9 (approved):
Great work, thank you :)

PR #319, opened 2021-12-28
Refactor/hadlers

genie9:
### Description

Refactors most handler classes to own files and updates tests to the new setup. 

### Type of change

Refactoring

### Changes Made

Separated most of the handlers to own files inside the handlers folder.
Refactored tests for a new setup.
Fixed imports in server.py.

### Testing

- [x] Unit Tests


metadata_backend/api/handlers/api_handlers.py
```diff
@@ -0,0 +1,458 @@
...
+        mimetypes.types_map[".js"] = "application/javascript"
+        mimetypes.types_map[".js.map"] = "application/json"
+        mimetypes.types_map[".svg"] = "image/svg+xml"
+        mimetypes.types_map[".css"] = "text/css"
+        mimetypes.types_map[".css.map"] = "application/json"
+        LOG.debug("static paths for SPA set.")
+        return self.path / "static"
+
+
+# Private functions shared between handlers
```
blankdots, line 412: for this i think we can create a separate file `common.py` in handlers submodule, makes the handlers cleaner and we will most likely add more. e.g. I will add more things with CSV PR.

tox.ini
```diff
@@ -50,6 +50,13 @@ deps =
     bandit
 commands = bandit -r metadata_backend/
 
+[testenv:documents]
```
blankdots, line 53: we already have `testenv:docs` maybe we can name it like this (to avoid confusion):
```suggestion
[testenv:spelling]
```

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py38, flake8, mypy, docs, black, bandit
+envlist = py38, flake8, mypy, docs, black, bandit, pyspelling
```
blankdots, line 2: the environment here needs to match the one below, if you put `documents` then this needs to be `documents` if we change to `spelling` then that it is.

blankdots (changes requested):
i think the naming scheme could be also simpler:
```
from .api.handlers.api
from .api.handlers.folder
from .api.handlers.object
from .api.handlers.submission
from .api.handlers.user
```

no need for the `_handler` at the end as we already have the module `handlers`

blankdots (approved):
looks way manageable now.

tox.ini
genie9: Reverted to original tox.in as it is not practical to configure for the local environment. To configure and run it in its own local environment instructions can be at found https://facelessuser.github.io/pyspelling/. I will add this to documentation.

Issue #335, opened 2022-01-21
Extend  object endpoint to handle patching


genie9:
The frontend is handling folder patching on an object POST by adding the newly created object to the user's folder. It would be better that the backend would handle the patching of an object as it is the default operation and the backend has all the data. This will also simplify integration tests.

Will introduce breaking changes to frontend:
- [ ] CSCfi/metadata-submitter-frontend#626

Steps for submitter API user:
1. user POST folder - to create new folder - notes the folder_id down
2. user does POST object - it provides also the folder_id in the request - for example: query string
3. user PATCH/PUT object - they don't need to know the folder_id - as the object is already in the folder

PR #341, opened 2022-01-27
Refactor/patch object on save

genie9:
### Description

Move responsibility of adding an object to a folder on object creation and update from frontend to object endpoint. 

###Related issues
Closes https://github.com/CSCfi/metadata-submitter/issues/335

### Type of change
Frontend now should add folder ID as a query parameter with object POST.

### Changes Made
Folder id is now required query param with object POST.
A folder is patched after an object is created and updated.
Adds extraction of title for folder's objects display title.

### Testing
- [x] Integration Tests

### DIFF
--- a/metadata_backend/api/handlers/object.py
+++ b/metadata_backend/api/handlers/object.py
@@ -109,7 +109,7 @@ class ObjectAPIHandler(RESTAPIHandler):
         operator: Union[Operator, XMLOperator]
         if req.content_type == "multipart/form-data":
             _only_xml = False if schema_type in _allowed_csv else True
-            files, cont_type = await multipart_content(req, extract_one=True, expect_xml=_only_xml)
+            files, cont_type, _ = await multipart_content(req, extract_one=True, expect_xml=_only_xml)
             if cont_type == "xml":
                 # from this tuple we only care about the content
                 # files should be of form (content, schema)
@@ -222,7 +222,7 @@ class ObjectAPIHandler(RESTAPIHandler):
         content: Union[Dict, str]
         operator: Union[Operator, XMLOperator]
         if req.content_type == "multipart/form-data":
-            files, _ = await multipart_content(req, extract_one=True, expect_xml=True)
+            files, _, _ = await multipart_content(req, extract_one=True, expect_xml=True)
             content, _ = files[0]
             operator = XMLOperator(db_client)
         else:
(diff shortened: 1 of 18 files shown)


blankdots:
how is this different than #340 ?

blankdots:
Requires https://github.com/CSCfi/metadata-submitter-frontend/issues/626 in front-end

metadata_backend/api/handlers/object.py
```diff
@@ -283,3 +305,43 @@ async def patch_object(self, req: Request) -> Response:
...
+
+        :param schema: schema of objects to be added to the folder
+        :param ids: object IDs
+        :param params: addidtional data required for db entry
+        :returns: list of patch operations
+        """
+        if not params.get("cont_type", None):
+            submission_type = "Form"
+        else:
+            submission_type = params["cont_type"].upper()
```
blankdots, line 320: realised that for this to work we need to add `CSV` to lines:
- https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/helpers/schemas/folders.json#L999
- https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/helpers/schemas/folders.json#L1037

blankdots (commented):
if we add the object in the folder when creating, should we still allow patching https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/api/handlers/folder.py#L38 `metadataObjects` and `drafts` when PATCH folder ?

metadata_backend/api/handlers/object.py
genie9: strangely it worked without schema changes.

genie9:
> how is this different than #340 ?

changed the branch name because realized that we are not patching on the update. This obviously closed the PR.

genie9:
> if we add the object in the folder when creating, should we still allow patching https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/api/handlers/folder.py#L38 `metadataObjects` and `drafts` when PATCH folder ?

If e.g. title of an object will change i think it should be updated to the folder `metadataObjects` also.

blankdots:
> If e.g. title of an object will change i think it should be updated to the folder `metadataObjects` and `drafts` also. And actualy this is not happening now.

Good point, I think front-end does a PATCH or PUT for that I don't remember we need to check.

blankdots (commented):
looks good.
please add changes to `CHANGELOG.md`

blankdots (approved):

PR #346, opened 2022-02-07
Feature/project ownership

### LINKED ISSUE #282: Folders can be owned by multiple users
#### Description

Currently only one user can own and publish a folder, however there is a need that multiple members of a research group can operate on a folder, as it involves members with different understandings of the datasets

It would be good to introduce the concept of a `group` or `project` to correspond to CSC naming scheme and instead of assigning folders, drafts to a user assign them to a group.

#### DoD (Definition of Done)

Folders can be modified and published by users of different research groups

#### Testing

Unit and integration testing

teemukataja:
### Description

Folders and templates were originally owned by users, now they will be owned by projects that users belong to, allowing multiple users to work on the same folders and templates at the same time.

### Related issues
Fixes #282 
Frontend issue https://github.com/CSCfi/metadata-submitter-frontend/issues/625

### Type of change

- [x] This change requires a documentation update (API changes)

### Changes Made

- deprecated `folders` and `templates` keys from user
- added new key `projects` to user
- added new key `projectId` to `folder` and `template` collections
- added new `project` collection
- new mandatory userinfo value from AAI `sdSubmitProjects`
- if user has no `sdSubmitProjects` they are redirected to `/noproject`
- new mandatory query parameter `projectId` for `GET /folders`
- new endpoint `GET /templates?projectId=...` that replaces `GET /users/current "templates":[...]`
- `projectId` is not mandatory when requesting **specific** resource, e.g. `GET /folders/{folderId}`, the permissions are read from session
- new mandatory key `projectId` for `POST /folders and /templates`
- deprecated `PATCH /user`
- new request body keys `index` and `tags` in `PATCH /templates/schema/templateId` which are the same values as previously used in the deprecated `PATCH /user`

#### Other Points
- user project affiliations are re-validated on every login, if project affiliations have changed, user data is updated
- breaking change that requires fresh database, because "project" is new information that did not exist before, and it can't be migrated to existing user-owned hierarchy due to missing relationship information

### Testing

- [x] Unit Tests
- [x] Integration Tests
- [x] Manual Testing


teemukataja:
The previous commit [00cec8f](https://github.com/CSCfi/metadata-submitter/pull/346/commits/00cec8fb4b6042c8c339066651a6ec670b99b3cc) updated integration tests, they also showed some bugs in the code which were fixed in the same commit. Also removed some deprecated features, and will continue to prune through deprecated features in next commit.

teemukataja:
Added integration test case for `GET /templates`. I tried to copy folder unit tests (`query_folders`) for templates (`query_templates_by_project`) (because they are almost identical), but because the query loops over collections, it raises a `StopAsyncIteration`, and I'm not sure how to deal with that.

genie9:
> `StopAsyncIteration`
 Would this be usful:
```    
try:
    ....    
except StopIteration:
    pass
```

teemukataja:
> > Added integration test case for `GET /templates`. I tried to copy folder unit tests (`query_folders`) for templates (`query_templates_by_project`) (because they are almost identical), but because the query loops over collections, it raises a `StopAsyncIteration`, and I'm not sure how to deal with that.
> 
> Would this be usful:
> 
> ```
> try:
>     ....    
> except StopIteration:
>     pass
> ```
> 
> https://github.com/CSCfi/metadata-submitter/blob/135af96bc22ce651019955356c97122a9746369f/tests/integration/run_tests.py#L549

I thought about this also initially and tested it, but it still errored, and I think it's also not the correct way to approach this, because it would stop the test

This is the function
https://github.com/CSCfi/metadata-submitter/blob/feature/project-ownership/metadata_backend/api/operators.py#L308-L365
It queries multiple collections

It's based on this
https://github.com/CSCfi/metadata-submitter/blob/feature/project-ownership/metadata_backend/api/operators.py#L812-L854

blankdots:
> It's based on this https://github.com/CSCfi/metadata-submitter/blob/feature/project-ownership/metadata_backend/api/operators.py#L812-L854

I think this: https://github.com/CSCfi/metadata-submitter/blob/feature/project-ownership/metadata_backend/api/operators.py#L395 might be more appropriate as you want object query not folder query and maybe pay attention to: https://github.com/CSCfi/metadata-submitter/blob/feature/project-ownership/metadata_backend/conf/conf.py#L113

blankdots:
wait you want `query_templates_by_project` not to filter the templates content ... :thinking: maybe it will  be easier if you put them all in one collection `templates` and add the type for each document in the collect e.g. study, sample etc. I am not sure we should invest the effort in making cross collection queries at this point for `templates`

blankdots:
however if you really want JOIN like from SQL i think https://docs.mongodb.com/manual/reference/operator/aggregation/lookup/ is what you need.

metadata_backend/api/handlers/object.py
```diff
@@ -101,6 +101,13 @@ async def post_object(self, req: Request) -> Response:
         """
         _allowed_csv = ["sample"]
         schema_type = req.match_info["schema"]
+
+        folder_id = req.query.get("folder", "")
```
blankdots, line 105: Not sure we do the check if the folder ID is owned by a project the user has access to, otherwise we risk that we get a folder ID from a project the user does not have access to, or do you think this case is plausible ?

probably in #354 we should have checked if it is owned by a user, but we can leave that be and correct it moving forward.

teemukataja: True, the project is checked in all other methods (GET, PUT, PATCH, DELETE) but not in POST, I'll add it

tests/test_handlers.py
```diff
@@ -824,8 +824,6 @@ async def test_get_folders_with_bad_params(self):
 
     async def test_get_folder_works(self):
         """Test folder is returned when correct folder id is given."""
-        RESTAPIHandler._handle_check_ownership = make_mocked_coro(True)
```
teemukataja: removed a duplicate row that seemed to do nothing, the function is already mocked in the test case setup https://github.com/CSCfi/metadata-submitter/blob/feature/project-ownership/tests/test_handlers.py#L103

teemukataja:
> > however if you really want JOIN like from SQL i think https://docs.mongodb.com/manual/reference/operator/aggregation/lookup/ is what you need.
> 
> I will look into this if it makes it possible to request the data in one query

Looks like lookup is used for joining data together into one object from separate collections, but we want to get a list of all objects from multiple collections that have the same project id

teemukataja:
Returned old object assigning code into use with templates-in-project (as opposed to templates-in-user).

Missing feature: `PATCH /template` requires feature for updating `templates`-key in `project` (similarly to `PATCH user templates`)

blankdots (approved):

PR #353, opened 2022-02-15
check values in set instead of list

blankdots:
### Description

checking a set should be faster than checking a list

### Type of change

- [x] Refactor

### Changes Made

1. check in set rather than a list, in an attempt to get every last drop of performance :)  

### Testing

- [x] Tests do not apply


blankdots:
going to merge this so it is easier to work with all the changes in #332

PR #354, opened 2022-02-16
Refactor/patch object on save

genie9:
### Description

Move responsibility of adding an object to a folder on object creation and update from frontend to object endpoint. 

###Related issues
Closes https://github.com/CSCfi/metadata-submitter/issues/335

### Type of change
Frontend now should add folder ID as a query parameter with object POST.

### Changes Made
Folder id is now required query param with object POST.
A folder is patched after an object is created and updated.
Adds extraction of title for folder's objects display title.

### Testing
- [x] Integration Tests

### DIFF
--- a/.github/config/.wordlist.txt
+++ b/.github/config/.wordlist.txt
@@ -115,6 +115,7 @@ cscfi
 cscusername
 csi
 csrf
+csv
 ctrl
 cts
 curation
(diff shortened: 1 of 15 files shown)


blankdots (approved):
seems front-end is ready with updates https://github.com/CSCfi/metadata-submitter-frontend/pull/672
 @genie9  can we merge this ?

PR #399, opened 2022-04-11
Fix/person required fields

genie9:
### Description
While publishing the submission in SD Submit, the request failed due to a missing field error coming from Metax. 
Fixed by adding `affiliation` as a required field for `creators` and `contributors` in Datacite schema. This field is needed to map `member_of`-field for `creator`, `curator`, `rights_holder` and `contributor` from Metax schema. 

### Type of change

### Changes Made

Adds `affiliation` to required fields for `creators` and `contributor` in Datacite schema

### Testing

- [x] Tests do not apply

### DIFF
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -453,7 +453,7 @@ class Operator(BaseOperator):
             LOG.error(reason)
             raise web.HTTPBadRequest(reason=reason)
         else:
-            LOG.info(f"Object {schema_type} with id {accession_id} opdated with metax info.")
+            LOG.info(f"Object {schema_type} with id {accession_id} metax info updated.")
             return True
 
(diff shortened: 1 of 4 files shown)


genie9:
> folder.json field `doiInfo` needs updating with the same information
> 
> of we can try to use JSON schema `$ref` maybe that will work nicely

I'll check that, thanks for pointing out.

blankdots (approved):
the datacite.json will be moved to folder.json as part of #400

genie9:
@blankdots is the last commit not needed based on your last PR? But I guess it will be just overwritten...

blankdots:
> @blankdots is the last commit not needed based on your last PR? But I guess it will be just overwritten...

we can keep it. I will rebase my PR to get the new changes

PR #421, opened 2022-05-10
Refactor/aiohttp session

genie9:
### Related issues

Introduce aiohttp_session as a way to handle the session instead of the custom way that was there before.

### Type of change

### Changes Made

1. added `aiohttp_session` as dependency
2. fixed function comments and missing information
3. refactored auth to include  new session management 
4. refactored unit tests
5. fixed some typing `Any`
6. migrated to variables used by motor 3 for SSL https://motor.readthedocs.io/en/stable/migrate-to-motor-3.html?highlight=ssl_certfile#renamed-uri-options

### Testing

- [x] Unit Tests


hannyle:
I tested this branch and e2e test still fail from github CI pipeline.

csc-felipe:
@hannyle, does it also fail if you run locally?

hannyle:
> @hannyle, does it also fail if you run locally?

They run fine locally.

genie9:
@hannyle is it the same redirect issue?

hannyle:
> @hannyle is it the same redirect issue?

Yes, the issue of redirection from `localhost:5430` --> `localhost:5430/error401`

csc-felipe:
We have changed so much code after this was initially made! I will try to make it work again, and see if the previous issues will still exist.

csc-felipe:
It seems to run as expected ^^. Still need to investigate the previously existing issues.

csc-felipe:
The issue seems to come from changing the setup for checking for valid sessions in `middlewares.py`. There used to be a check for static files, which was removed. Restoring that removes the unexpected redirect.

It's still a bit of an issue, though. Because the previous code was checking for static routes, which the new one discarded that. But at the same time, we don't know all the static routes that exist, so we can't check them. We should instead serve static files only with external hosting, or change the setup.

We don't really test static routes in the backend, but the frontend does.

Checking sessions was done on a global level, but it shouldn't check routes that match the static routes that would match the frontend. Those should always return the static files.

It would be nice to be able to mark (decorate) routes that are checked, instead of a global check which should keep track of which routes exist. Sounds like a prone to error approach.

Perhaps we should use nested applications, so we can apply the session checking middleware only to the API routes, and the public routes are not checked. We might need then to prepend the API routes with, for e.g. `/api`. I think this would be a good solution.

csc-felipe:
This PR was already quite large, so I'll make another PR, which makes some changes to API, and fixes the issues created by this one, mentioned above.

teemukataja (approved):
with https://github.com/CSCfi/metadata-submitter/pull/476

Issue #382, opened 2022-03-28
Metax error handling backend/frontend


genie9:
There needs to be a way to identify where from error is coming, is it Metax or an internal error.
One idea would be to add to errors coming from Metax service a suffix indicating that this is an error from an external service. Also maybe some informative text on how to deal with this error.

PR #433, opened 2022-05-18
HTTPError exceptions return a JSON Problem instead of HTML.

csc-felipe:
### Description

The backend should remain an API, and as such, return errors as JSON, and not HTML.
This change removes HTML file responses for HTTP errors, and uses JSON Problem detail responses instead.

### Type of change

### Changes Made

- Refactor HTTPError responses to return a JSON Problem detail object.
- Update JSON Problem detail to include HTTP status code
- Update OpenApi Swagger docs

### Testing

- [x] Tests do not apply


metadata_backend/api/middlewares.py
```diff
@@ -43,32 +28,20 @@ async def http_error_handler(req: Request, handler: Callable) -> Response:
         response = await handler(req)
         return response
     except web.HTTPError as error:
-        details = _json_exception(error.status, error, req.url)
-        LOG.error(details)
+        LOG.error(error)
+        problem = _json_problem(error, req.url)
+        LOG.info(problem)
```
teemukataja, line 33: Should we add some description to the logs, and not only the variables?

teemukataja (approved):
Small comment on logging

metadata_backend/api/middlewares.py
csc-felipe: I've been thinking about it, and it's somehow in the back of my mind, but I didn't come up with a solution.

csc-felipe:
@teemukataja, I think logging is better now, what do you think?

teemukataja (approved):

genie9:
There is a need to address how Metax errors would be recognized from submitter errors as this PR is marked to solve #382. We will discuss that with @csc-felipe

genie9 (changes requested):
https://github.com/CSCfi/metadata-submitter/pull/433#issuecomment-1134492028

csc-felipe:
> There is a need to address how Metax errors would be recognized from Submitter errors as this PR is marked to solve #382. We will discuss that with @csc-felipe

Makes. But I don't really understand what does it mean. Let's merge this PR, if it's desired, or close it if not, and then the other change comes in a new PR.

genie9 (approved):

csc-felipe:
> The JSON errors do not cause any breaking changes in frontend except in some cases I think it's better if the backend can still redirect to html page instead of a JSON format. For example, when the user makes a page request and receives 400 or 401.

@hannyle, Because we are offering the API as a service to customers, it's more natural to give a JSON response than a redirect. Would it be ok if the frontend then redirects to the error page in the front end?

hannyle:
> > The JSON errors do not cause any breaking changes in frontend except in some cases I think it's better if the backend can still redirect to html page instead of a JSON format. For example, when the user makes a page request and receives 400 or 401.
> 
> @hannyle, Because we are offering the API as a service to customers, it's more natural to give a JSON response than a redirect. Would it be ok if the frontend then redirects to the error page in the front end?

In the old way, I think the backend also returned JSON error for API requests and the frontend is already configured that it would intercept and transform the error to either a redirection or something else. 

The issue I mentioned is not with API requests but for cases that user types a wrong URL that would return 400 or 401, and with that the frontend currently cannot intercept in the same way as API requests. With the old backend, it would return html page instead, but if this function is removed I would need to figure out another way to configure it frontend, I don't know how yet it may take sometime.

csc-felipe:
> The issue I mentioned is not with API requests but for cases that user types a wrong URL that would return 400 or 401

Now I get it. I'm working on another change #421, so that it will always return the frontend HTML in that case, instead of an error (as part of fixing that PR). So you can do what you want with the wrong URL.

I'll merge this PR.

hannyle:
> Now I get it. I'm working on another change https://github.com/CSCfi/metadata-submitter/pull/421, so that it will always return the frontend HTML in that case, instead of an error (as part of fixing that PR). So you can do what you want with the wrong URL.

> I'll merge this PR.

That sounds good to me! Please go ahead :)

csc-felipe:
The fix actually went to PR #476.

Issue #430, opened 2022-05-17
Investigate use of submission patch and refactor it to meet the needs


genie9:
#### Description
The use of endpoint for patching submission has changed radically during a few code updates. Currently, it is probably used to only update the submission's `name` and `description` but it needs to be investigated. If this is the case, then the patching function should be minimized to accept only those fields as JSON and no extra db-related information.
Current request body for the call:
` [{"op": "replace", "path": "/name", "value": "test2"}]`
Expected request body for the call:
`{"name": "test2"}`

#### DoD (Definition of Done)
Call to patch `submission/<submission_id>` will update only submission's `name` and `description` with simple JSON string.
The frontend has been adapted with this update.
Tests and documentation are updated with changes if needed.

#### Testing
Integration tests both on the front and the backend

csc-felipe:
Replaced **folder** with **submission**.

csc-felipe:
The other endpoints still support PATCH with the same syntax. Should we only restrict which fields in submission can be updated with PATCH, or also refactor other PATCH endpoints to function similarly, without the JSON:patch syntax?

I think it's worse UX that submission PATCH endpoint has different format than the other PATCH endpoints.

csc-felipe:
This change also means that some possible use cases would be removed, for e.g., what's now in integration tests: https://github.com/CSCfi/metadata-submitter/blob/53c60c7e6242ff3793427ee2f9edc7895bcf70f7/tests/integration/run_tests.py#L1606=

PR #450, opened 2022-05-25
HTTP PATCH for submissions only accepts 'name' and 'description'.

csc-felipe:
### Description

We're deprecating the use of JSON Patch operations to modify most properties of the submission. Now only 'name' and 'description' can be updated, and in a flat JSON object, example: `{"name": ""}`.

### Related issues

- Closes https://github.com/CSCfi/metadata-submitter/issues/430

### Type of change

### Changes Made

- Submission HTTP PATCH doesn't accept a JSON Patch any more, only specific fields.
- Updated unit and integration tests
- Updated swagger docs

### Testing

- [x] Unit Tests
- [x] Integration Tests


hannyle:
Frontend modification for this has passed the tests and in review.
I think it's safe to merge this PR to `develop` if you would like to proceed.

genie9 (dismissed):
Looks really good! The only missing part is the swagger documentation on submission patching.
The example of the patch needs to be updated.

genie9 (approved):

PR #473, opened 2022-06-06
Fix XML parsing error

csc-jm:
### Type of change

### Changes Made

-changed faulty parsing method

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Tests do not apply


csc-felipe (changes requested):
Nice find and fix! Could you add some tests too? So we don't make regressions on it again.

genie9 (approved):
As Felipe said updating the test for this case would be a wanted addition. Otherwise good fix!

Issue #448, opened 2022-05-25
Make integration tests idempotent


csc-felipe:
#### Proposed solution

Currently, the integration tests modify the DB state, and depend on each other.

It makes it difficult to develop new tests, as DB must be cleared manually after every run.

#### DoD (Definition of Done)

- Integration tests can be run multiple times in a row.
- Failing tests leave the environment in a clean state.

#### Testing

Run integration tests multiple times in the same environment, and observe the same output / result.

csc-felipe:
Perhaps we could use [pytest with fixtures for this](https://docs.pytest.org/en/6.2.x/fixture.html#fixtures).

blankdots:
> Currently, the integration tests modify the DB state, and depend on each other.

could we get more info on this, as i remember they were kind self contained each testing a different thing and creating its own fixtures?

they do leave a lot of `submissions` just to identify there aren't any interactions between each `submission` data as we have had concurrency issues in the past.

the solution done by #477 should have fixed this ?

csc-felipe:
I think that PR did the most important part, which was to be able to re-run tests without intermediate commands. It does still bother me, that the development DB is the same as the test DB though.

What I was having in mind next was to migrate the tests to use Pytest and fixtures. I think it will allow to more easily create certain kinds of tests, for e.g. no metax service running, and such. Should we have a chat about this?

csc-felipe:
#570 was a step forward, but I would still like to 
- split integration tests into multiple files,
- be able to control running services from the code (to test for when a service is down, for example).

PR #477, opened 2022-06-08
Recreate db schema before integration tests.

csc-felipe:
### Description

Hopefully this makes life a bit easier with integration tests. I was running them so many times while debugging for another change, that it felt like the right time to do this.

### Related issues

- Related to #448 

### Type of change

### Changes Made

- Refactored `mongo_indexes.py` to be useful
- Integration tests always recreate db at start, and clean it up afterwards

### Testing

- [x] Tests do not apply


csc-jm (approved):
This seems like a good change to me

PR #480, opened 2022-06-10
Make Metax integration optional

### LINKED ISSUE #467: Make using of discovery service (METAX) optional
#### Description

We have been implementing integration with the FAIRDATA Metax service as part of the standard workflow for submission publishing.
Now we have a need to make that integration optional as BIGPICTURE submissions won't be using FAIRDATA as a discovery service.
The proposition is to add an environment variable that would indicate what workflow will be used. This env var will probably be used also for schema validation so it could be something like `SUBMIT_WORKFLOW` and value `bigpicture`. 
As for FAIRDATA `SUBMIT_WORKFLOW` value could be `standard` as it is considered a standard use case.

#### DoD (Definition of Done)

When submitting metadata and publishing submission with env var `SUBMIT_WORKFLOW: bigpicture` submissions' metadata is not sent to Metax.

#### Testing

Not yet decided, but coming soon...

csc-felipe:
### Description

Make Metax integration optional, so that new use cases can be supported.

### Related issues

- Closes #467
- Follow-up on #512 
- Follow-up on #513
### Type of change

It is actually breaking, because it needs configuration update in the deployment, adding environment variable to enable the integration `METAX_ENABLED=True`. It doesn't break the frontend though.

### Changes Made

- New environment variable to indicate the use of Metax integration.
- Surround metax calls with a check if metax is enabled.

### Testing

- [x] Needs testing (start an issue or do a follow up PR about it) #513

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -35,6 +35,8 @@ DOI_KEY=key
 DISCOVERY_URL=https://etsin.demo.fairdata.fi/dataset/
 
 # metax
+# Must be 'True' to enable
+METAX_ENABLED=True
 METAX_USER=sd
 METAX_PASS=demo_pass
 METAX_URL=http://mockmetax:8002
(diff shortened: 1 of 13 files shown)


genie9 (commented):
Here https://github.com/CSCfi/metadata-submitter/blob/b30558eb488ca850ab695886ddbf2fb0ffb9f8e9/metadata_backend/api/handlers/xml_submission.py#L256 is missing chack for `metax_anabled`

genie9 (approved):
GJ!

PR #491, opened 2022-06-13
Add support for posting BP samples

### LINKED ISSUE #490: Add BP.sample.xsd schema validation to Metadata Submitter API
#### Description

We need to adapt the metadata submitted to accept a BP Sample schema so corresponding XML files can be submitted.

#### DoD (Definition of Done)

A BP sample XML test file is validated against the corresponding schema and metadata is added to the database.

#### Testing

Unit tests

genie9:
### Description
Add schema validation for BP sample.

### Related issues
Closes #490 

### Type of change

### Changes Made

- Added new xml and json schema `bpsample` for validation and database writes
- Added extra method to `XMLToJSONParser` to organize BP sample objects into one sample object
- Added couple tests and a test file with an example BP sample submission
### Testing

- [x] Unit Tests
- [x] Integration Tests

### DIFF
--- a/tests/integration/run_tests.py
+++ b/tests/integration/run_tests.py
@@ -750,6 +750,20 @@ async def test_crud_with_multi_xml(sess, submission_id):
             LOG.debug(f"Checking that {item['accessionId']} XML is in {_schema}")
             assert resp.status == 200, f"HTTP Status code error, got {resp.status}"
 
+    _schema = "bpsample"
+    _filename = "template_samples.xml"
+    data = await post_multi_object(sess, _schema, submission_id, _filename)
+    for item in data:
+        items.append(item)
+        async with sess.get(f"{objects_url}/{_schema}/{item['accessionId']}") as resp:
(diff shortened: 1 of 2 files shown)


csc-jm:
The BP sample structure always seemed odd since the `<SAMPLE_SET>` as seen in `template_samples.xml` includes separate elements `<BIOLOGICAL_BEING>`, `<SPECIMEN>`, `<BLOCK>`, `<SLIDE>` and not just everything in a simple `<SAMPLE>` tag. And then a slide refers to a block, block refers to a specimen and a specimen refers to a biological being. 

Since my understanding was that the objects from those elements related to each other would be considered as one "sample", I made a method to sort them in that way, so that we can just add them to the db under `bpsample` collection. Hence, the implemented method creates 2 `bpsample` objects from the `template_samples.xml`.

csc-felipe (approved):
I don't fully understand this yet, but looks alright.

Issue #327, opened 2022-01-10
Map schemas between Metax Dataset schema and SD Submit schemas Datacite, Study and Dataset


genie9:
#### Description
Metadata-submitter's Datacite, Study and Dataset schemas need to be mapped to  Metax Dataset schema:  https://raw.githubusercontent.com/CSCfi/metax-api/master/src/metax_api/api/rest/v2/schemas/att_dataset_schema.json.

#### Tasks
- [x] #505
- [x] #506
- [x] #507

#### Testing

genie9:
Had a conversation on possible field mappings with @heikkil, @fmorelloCSC, and @blankdots. 

#### Already mapped fields:
| SD Submit datacite or object | Metax research_dataset |
| -------------   | -------------------------------- |
| DOI            | preferred_identifier |
| title (object) | title |
| description (dataset) | description |
| abstract (study) | description |
| "CSC Sensitive Data Services for Research" | publisher |
| creators | creator |
| "restricted" | access_rights |

#### Will be mapped with this ticket:
| SD Submit datacite or object| Metax research_dataset |
| ------------- | -------------------------------- |
dates Modified | modified
dates Issued | issued
dates Collected | temporal
keywords | keyword
alternateIdentifiers | other_identifier
contributors (other than Rights Holder,  Data Curator, Distributor) | contributor
contributors Rights Holder  | rights_holder
contributors Data Curator | curator
type (study/dataset) | theme
language | language
geoLocations | spatial
sizes (MUST standardize to bytes) | total_remote_resources_byte_size

_Need some clarification:_

Need to consult with Yrjö Leino if these fields could be mapped or should we add an extra field to datacite schema
SD Submit datacite _subject_ ->  Metax research_dataset _field_of_science_

Will this field be describing SD Submit Study where described SD Submit Dataset belongs to?
"is_output_of": {
    "title": "Producer project",
    "description": "A project that has caused the dataset to be created",
    "$ref": "#/definitions/Project"
    }

And vice verse this field will be describing SD Submit Dataset which belongs to SD Submit Study being described?
"relation": {
    "@id": "http://purl.org/dc/terms/relation",
    "title": "Relation",
    "description": "A related dataset or other entity",
    }

Could this be link to REMS?
remote_resources{
    "title": "Remote resources",
    "description": "A concrete storage or expression format for the data in the dataset, for example a file, a database or a query interface to the data.",
    "$ref": "#/definitions/WebResource"
    }

[...]

PR #499, opened 2022-06-21
Feature/map relations metadata for metax

genie9:
### Description

Adds mapping for study and dataset relations: 
- study is referenced in datasets with `is_output_of` -field
- datasets are referenced in study with `relation` -field

### Related issues

Related #327 
### Type of change

### Testing

- [ ] Unit Tests
- [x] Integration Tests
- [ ] Tests do not apply

### DIFF
--- a/metadata_backend/api/handlers/submission.py
+++ b/metadata_backend/api/handlers/submission.py
@@ -216,7 +216,13 @@ class SubmissionAPIHandler(RESTAPIHandler):
                                 req, "study", study_data, create_draft_doi=False
                             )
 
-                        metax_ids.append({"doi": study_data["doi"], "metaxIdentifier": study_data["metaxIdentifier"]})
+                        metax_ids.append(
+                            {
+                                "schema": "study",
+                                "doi": study_data["doi"],
+                                "metaxIdentifier": study_data["metaxIdentifier"],
+                            }
+                        )
 
                         # there are cases where datasets are added first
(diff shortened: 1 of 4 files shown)


metadata_backend/api/handlers/submission.py
```diff
@@ -541,15 +560,19 @@ async def publish_submission(self, req: Request) -> Response:
...
+        )
+        await metax_handler.publish_dataset(metax_ids)
 
-        # Create draft DOI ??? and delete draft objects from the submission
+        # Delete draft objects from the submission
         for obj in submission["drafts"]:
             await obj_ops.delete_metadata_object(obj["schema"], obj["accessionId"])
 
-        # update study to metax with data comming from doi info
-        await metax_handler.update_dataset_with_doi_info(submission["doiInfo"], metax_ids)
```
csc-felipe, line 550: Why did you remove this line?

genie9: It was raised before the publication procedure starts to line 567, so nothing will be deleted before we know that metax is ok. Also had to extra DB call because data needed for relation mapping is added to `extra_info` -field

genie9:
Integration tests and production deployment test has not been run for some reason. Need to investigate

csc-felipe:
Now it's running the other tests.

genie9:
Unfortunately not, I will update documentation and we probably need to have a chat on the subject.

PR #514, opened 2022-06-29
Add mapping for language

### LINKED ISSUE #505: Map languages between Metadata-submitter and Metax

genie9:
### Description

Adds mapping of languages between submitter and metax. 

### Related issues

Closes #505 

### Type of change

### Changes Made

- Adds parsing of language data from lexovo and iata endpoints to dictionary needed for mapping data
- Adds mapping for languages
- Running with Metax integration requires generating updated references file
- Docker builds automatically generates the required file

### Testing

- [x] Integration Tests


csc-felipe (changes requested):
That's a nice solution!

I think it would be nicer if the language mapping would be generated with a separate script, for example when building the docker image, or even included in the repo, but so that it's not generated at runtime. 
And it can be updated with a script. So it will be in its own python file with all the data generated or a JSON file, like the `identifier_types.json`.

This is mainly, so we don't depend on an external service to be available to run / test / debug the submitter.

It would also be good to log the KeyErrors.

genie9:
@csc-felipe could make Docker-dev generate metax_reference files. Script is running but nothing is added to the folder :(

csc-felipe:
Integration tests run with `Dockerfile-dev`, and files must be generated before installing the package. I was all the time looking at the `Dockerfile` instead. Now it works :)

teemukataja (approved):

csc-jm (approved):
Seems good

PR #525, opened 2022-07-07
Process multiple objects from same XML file

### LINKED ISSUE #479: XML submissions cannot process multiple objects from same file
#### Describe the bug

All test xml files at the moment have only one child within set. E.g.:
```
<?xml version = '1.0' encoding = 'UTF-8'?>
<POLICY_SET>
    <POLICY alias="asgsasg" center_name="TEST" broker_name="EGA" accession="EGAP50003522">
        <TITLE>Some policy</TITLE>
        <DAC_REF accession="EGAC500003259"/>
        <POLICY_TEXT>Do this, don't do that</POLICY_TEXT>
        <DATA_USES>
            <DATA_USE ontology="test" code="test2" version="1"/>
            <DATA_USE ontology="test1" code="test3" version="2"/>
        </DATA_USES>
    </POLICY>
</POLICY_SET>
```

When this is parsed into json, you get one `policy` object. However, what if the xml had for example this structure:
```
<?xml version = '1.0' encoding = 'UTF-8'?>
<POLICY_SET>
    <POLICY alias="asgsasg" center_name="TEST" broker_name="EGA" accession="EGAP50003522">
        <TITLE>Some policy</TITLE>
        <DAC_REF accession="EGAC500003259"/>
        <POLICY_TEXT>Do this, don't do that</POLICY_TEXT>
        <DATA_USES>
            <DATA_USE ontology="test" code="test2" version="1"/>
            <DATA_USE ontology="test1" code="test3" version="2"/>
        </DATA_USES>
    </POLICY>
    <POLICY alias="asgsasgaSas" center_name="TEST1" broker_name="EGA" accession="EGAP50003522">
        <TITLE>Some other policy</TITLE>
        <DAC_REF accession="EGAC500003259"/>
        <POLICY_TEXT>Do this, don't do that</POLICY_TEXT>
        <DATA_USES>
            <DATA_USE ontology="test" code="test2" version="1"/>
            <DATA_USE ontology="test1" code="test3" version="2"/>
        </DATA_USES>
    </POLICY>
</POLICY_SET>
```

The `XMLToJSONParser()` makes a list of policies instead of a single dictionary and the list cannot be further validated and added to DB.

#### To Reproduce
Steps to reproduce the behaviour:
1. Change some xml test file to have multiple objects in a set (like shown with policy above)
2. Add the file to `test_xml_files` list in `tests/integration/run_tests.py`
3. Run the integration tests (`test_crud_works()` will not pass)

#### Expected behavior

When XML metadata files have multiple children at the base level, those should be able to be submitted to our database.

#### Additional context

Found during integrating BP xml files

csc-jm:
### Description

This PR implements a change to the `post_object()` method so that if an XML file contains multiple children under the main element (see the `policy2.xml` file for an example with 2 `<POLICY>` elements under `<POLICY_SET>`), it parses, validates and adds the objects to the database separately in the same request and returns with a single response. 

This shouldn't be a breaking change, since submitting single objects via XML still functions and returns with same responses as before. Also in the current state it seems to be possible to submit multiple objects from separate XML files at once so this change wasn't that big of a refactor. One thing to note however is that while each object gets added separately in JSON format, the entire XML is saved as text under the same accession ID in XML format. Not the smartest thing but probably fine for now since we don't do much with the XMLs after they've been parsed to JSON. 

### Related issues

Fixes #479 

### Type of change

### Changes Made

- Changed the methods that extract and post objects from xml files so that mutiple objects can be posted from the same file in the same request
- Added some new and altered some previous unit tests
- Added a test file with that has two policies within the policy set element
- Added new integration tests for the multi object submission from single files 
- Fixed some docstrings

### Testing

- [x] Unit Tests
- [x] Integration Tests

### DIFF
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -469,7 +469,7 @@ class Operator(BaseOperator):
 
         :param schema_type: Schema type of the object to create.
         :param data: Metadata object
-        :returns: Tuple of Accession Id for object inserted to database and its title
+        :returns: Metadata object with some additional keys/values
         """
         accession_id = self._generate_accession_id()
         data["accessionId"] = accession_id
@@ -495,7 +495,7 @@ class Operator(BaseOperator):
         :param schema_type: Schema type of the object to replace.
         :param accession_id: Identifier of object to replace.
         :param data: Metadata object
-        :returns: Tuple of Accession Id for object replaced in database and its title
+        :returns: Metadata object with some additional keys/values
         """
         forbidden_keys = {"accessionId", "publishDate", "dateCreated", "metaxIdentifier", "doi"}
         if any(i in data for i in forbidden_keys):
@@ -606,7 +606,7 @@ class XMLOperator(BaseOperator):
 
         :param schema_type: Schema type of the object to read.
         :param data: Original XML content
-        :returns: Tuple of Accession Id for object inserted to database and its title
+        :returns: List of metadata objects extracted from the XML content
         """
         db_client = self.db_service.db_client
         # remove `draft-` from schema type
@@ -635,7 +635,7 @@ class XMLOperator(BaseOperator):
         :param schema_type: Schema type of the object to replace.
         :param accession_id: Identifier of object to replace.
         :param data: Original XML content
-        :returns: Accession Id for object inserted to database
+        :returns: Metadata object extracted from the XML content
         """
         db_client = self.db_service.db_client
         # remove `draft-` from schema type
--- a/metadata_backend/api/operators.py
+++ b/metadata_backend/api/operators.py
@@ -39,7 +39,7 @@ class BaseOperator(ABC):
         self.db_service = DBService(db_name, db_client)
         self.content_type = content_type
 
-    async def create_metadata_object(self, schema_type: str, data: Union[Dict, str]) -> Dict:
+    async def create_metadata_object(self, schema_type: str, data: Union[Dict, str]) -> Any:
         """Create new metadata object to database.
 
         Data formatting and addition step for JSON or XML must be implemented
@@ -47,11 +47,11 @@ class BaseOperator(ABC):
 
         :param schema_type: Schema type of the object to create.
         :param data: Data to be saved to database.
-        :returns: Tuple of Accession id for the object inserted to database and its title
+        :returns: Dict of Accession id for the object inserted to database and its title
         """
         data = await self._format_data_to_create_and_add_to_db(schema_type, data)
-        data_list = data if isinstance(data, list) else [data]
-        acc_ids: list = []
+        data_list: List = data if isinstance(data, list) else [data]
+        acc_ids = []
         for object in data_list:
             acc_ids.append(object["accessionId"])
         LOG.info(f"Inserting object with schema {schema_type} to database succeeded with accession id(s): {acc_ids}")
@@ -254,7 +254,7 @@ class BaseOperator(ABC):
             raise web.HTTPNotFound(reason=reason)
 
     @abstractmethod
-    async def _format_data_to_create_and_add_to_db(self, schema_type: str, data: Any) -> Dict:
+    async def _format_data_to_create_and_add_to_db(self, schema_type: str, data: Any) -> Any:
         """Format and add data to database.
 
         Must be implemented by subclass.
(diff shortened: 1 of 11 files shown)


metadata_backend/api/operators.py
```diff
@@ -39,20 +39,22 @@ def __init__(self, db_name: str, content_type: str, db_client: AsyncIOMotorClien
         self.db_service = DBService(db_name, db_client)
         self.content_type = content_type
 
-    async def create_metadata_object(self, schema_type: str, data: Union[Dict, str]) -> Dict:
+    async def create_metadata_object(self, schema_type: str, data: Union[Dict, str]) -> Any:
```
csc-felipe, line 42: Could the return type be more precise?

metadata_backend/api/operators.py
```diff
@@ -39,20 +39,22 @@ def __init__(self, db_name: str, content_type: str, db_client: AsyncIOMotorClien
...
         """Create new metadata object to database.
 
         Data formatting and addition step for JSON or XML must be implemented
         by corresponding subclass.
 
         :param schema_type: Schema type of the object to create.
         :param data: Data to be saved to database.
-        :returns: Tuple of Accession id for the object inserted to database and its title
+        :returns: Dict of Accession id for the object inserted to database and its title
```
csc-felipe, line 50: Nice if this matches the return type.

metadata_backend/api/operators.py
```diff
@@ -252,7 +254,7 @@ async def check_exists(self, schema_type: str, accession_id: str) -> None:
             raise web.HTTPNotFound(reason=reason)
 
     @abstractmethod
-    async def _format_data_to_create_and_add_to_db(self, schema_type: str, data: Any) -> Dict:
+    async def _format_data_to_create_and_add_to_db(self, schema_type: str, data: Any) -> Any:
```
csc-felipe, line 257: Could the return type be more precise?

metadata_backend/api/operators.py
```diff
@@ -597,7 +599,7 @@ def __init__(self, db_client: AsyncIOMotorClient) -> None:
         """
         super().__init__(mongo_database, "text/xml", db_client)
 
-    async def _format_data_to_create_and_add_to_db(self, schema_type: str, data: str) -> Dict:
+    async def _format_data_to_create_and_add_to_db(self, schema_type: str, data: str) -> List:
```
csc-felipe, line 602: When typing a list, it's nice to say what are the contents: `List[dict]` or is it `List[tuple]`?

It would also be nice if the doc string is matching the return type.

metadata_backend/api/operators.py
```diff
@@ -39,20 +39,22 @@ def __init__(self, db_name: str, content_type: str, db_client: AsyncIOMotorClien
...
         :param data: Data to be saved to database.
-        :returns: Tuple of Accession id for the object inserted to database and its title
+        :returns: Dict of Accession id for the object inserted to database and its title
         """
         data = await self._format_data_to_create_and_add_to_db(schema_type, data)
-        LOG.info(
-            f"Inserting object with schema {schema_type} to database succeeded with accession id: {data['accessionId']}"
-        )
+        data_list: List = data if isinstance(data, list) else [data]
+        acc_ids = []
```
csc-felipe, line 54: It looks like nothing is done with this. I guess you intended to return it?

csc-felipe (changes requested):
This looks nice. Left a few comments about the typing. 

It would be nice if we don't have to use `Any` as return type.

metadata_backend/api/operators.py
csc-jm: Yes it can. Resolved

metadata_backend/api/operators.py
csc-jm: Actually noticed many of these were wrong and tried to fix them

csc-jm:
Typing is now more accurate without refactoring/creating separate methods for json and xml object parsing/creation. Also noticed some docstring return types were wrong so updated those. I wonder however, if `put_object()` method needs to take XML's with multiple objects into account...

metadata_backend/api/operators.py
teemukataja: stale

teemukataja (approved):

PR #530, opened 2022-07-11
Refactor api handlers to share service handler instances

csc-felipe:
### Description

I made this as part of implementing the Rems integration #498. But thought it's better to split into a separate PR.

Makes it more clear where calls to integration services are coming
from, improving maintainability of the code.

Mocking Datacite and Metax is not necessary in places where errors are ignored (most object endpoints). Otherwise, they are mocked where needed. Thus becoming more clear.

I wonder if the changes to unit tests are actually good overwall. At least for me it makes things easier to follow and understand what's happening.

### Type of change

### Changes Made

- Reuse instance of service handlers.
- Properly close HTTP clients.

### Testing

- [x] Tests do not apply

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Refactor authentication checking, fixing issues from #421 and remove HTTPSeeOther from the API
 - Recreate DB before integration tests run and cleanup after integration tests have run #448
 - Make using of discovery service (METAX) optional #467
+- Refactor api handlers to share a single instance of service handlers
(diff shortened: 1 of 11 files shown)


teemukataja (approved):
separating integrations to their own class sounds like a good idea :+1:

csc-jm (approved):
Yea changes here seem good, although you could have merged this to that #498 branch instead so we could see how these changes go along with that PR.

csc-felipe:
> Yea changes here seem good, although you could have merged this to that #498 branch instead so we could see how these changes go along with that PR.

The other PR (#498) is already rebased on this branch, so you could already check how they work together ^^

PR #543, opened 2022-07-25
Update deps

csc-felipe:
`xmlschema` had a breaking change, so some refactoring was necessary.

Bandit flagged the solution of importing `ElementTree` from `xml.etree` as an issue, and `defusedxml` is supposed to be the answer.

### DIFF
--- a/metadata_backend/helpers/validator.py
+++ b/metadata_backend/helpers/validator.py
@@ -2,15 +2,16 @@
 
 import re
 from io import StringIO
-from typing import Any, Dict, cast
+from typing import Any, Dict
 from urllib.error import URLError
 
 import ujson
 from aiohttp import web
+from defusedxml.ElementTree import ParseError
+from defusedxml.ElementTree import tostring as etree_tostring
 from jsonschema import Draft7Validator, validators
 from jsonschema.exceptions import ValidationError
 from xmlschema import XMLSchema, XMLSchemaValidationError
-from xmlschema.etree import ElementTree, ParseError
 
 from ..helpers.logger import LOG
 from .schema_loader import JSONSchemaLoader, SchemaNotFoundException
@@ -23,7 +24,7 @@ class XMLValidator:
         """Set variables.
 
         :param schema: Schema to be used
-        :param content: Content of XML file to be validated
+        :param xml: Content of XML file to be validated
         """
         self.schema = schema
         self.xml_content = xml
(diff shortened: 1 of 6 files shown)


metadata_backend/helpers/validator.py
```diff
@@ -23,7 +24,7 @@ def __init__(self, schema: XMLSchema, xml: str) -> None:
         """Set variables.
 
         :param schema: Schema to be used
-        :param content: Content of XML file to be validated
+        :param xml: Content of XML file to be validated
```
csc-jm, line 27: Nice thing that code style checks miss these things... :D

csc-jm (approved):
Seems good

teemukataja (approved):

metadata_backend/helpers/validator.py
csc-felipe: Yea, it's `pylint` that checks these things. I'm trying to configure it :)

metadata_backend/helpers/validator.py
```diff
@@ -2,15 +2,16 @@
...
 import re
 from io import StringIO
-from typing import Any, Dict, cast
+from typing import Any, Dict
 from urllib.error import URLError
 
 import ujson
 from aiohttp import web
+from defusedxml.ElementTree import ParseError
+from defusedxml.ElementTree import tostring as etree_tostring
```
blankdots, line 11: i don't think we needed to introduce here a new library i see xmlschema using: https://github.com/sissaschool/xmlschema/blob/master/xmlschema/__init__.py#L10

metadata_backend/helpers/validator.py
```diff
@@ -2,15 +2,16 @@
 
 import re
 from io import StringIO
-from typing import Any, Dict, cast
+from typing import Any, Dict
 from urllib.error import URLError
 
 import ujson
 from aiohttp import web
+from defusedxml.ElementTree import ParseError
```
blankdots, line 10: the same with this: https://github.com/sissaschool/xmlschema/blob/master/xmlschema/validators/schemas.py#L29

csc-felipe:
There was a change in `xmlschema` from release 2.0.0 https://github.com/sissaschool/xmlschema/releases/tag/v2.0.0. In which they refactored, and removed some code, which we were previously importing.

Then Bandit was complaining that directly using the other solution I used, (which I don't remember exactly what it was, but likely `xml.etree.ElementTree.parse`), to process user data was unsafe, with the suggestion to instead use `defusedxml` https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html?highlight=defusedxml#b313-b320-xml

csc-felipe:
I'll try to remember to provide the full info with references in the PR description. Ah, the commit message had some more context as well.

PR #544, opened 2022-07-25
Dependabot weekly -> monthly

csc-felipe:
Weekly updates are a bit too much for the size of our team, and frequency of releases. Let's do it monthly instead.


csc-jm (approved):
Sure

teemukataja (approved):

csc-felipe:
How can I fix tests? '^^

PR #546, opened 2022-07-27
Add Pylint checks

csc-felipe:
### Description

Using Pylint to check the code makes us a bit less prone to introduce technical debts unknowingly in the code.

Some checks have been disabled, and we may consider enabling and fixing some of them in the future.

### Type of change

### Changes Made

- Add checks with pylint
- Fix many pylint complaints
- Removed pydocstyle, which is already integrated into flake8
- I didn't see the benefit of `doc8` at this point, so removed it

### Testing

- [x] Tests do not apply

### DIFF
--- a/metadata_backend/api/auth.py
+++ b/metadata_backend/api/auth.py
@@ -45,10 +45,10 @@ class AccessHandler:
         }
         self.rph = RPHandler(self.oidc_url, client_configs=self.oidc_conf)
 
-    async def login(self, req: Request) -> Response:
+    async def login(self, _: Request) -> web.HTTPSeeOther:
         """Redirect user to AAI login.
 
-        :param req: A HTTP request instance (unused)
+        :param _: A HTTP request instance (unused)
         :raises: HTTPInternalServerError if OIDC configuration init failed
         :returns: HTTPSseeOther redirect to AAI
         """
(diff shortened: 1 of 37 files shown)


csc-jm (approved):
Initially I thought the list of checks in the pylint config file was a bit excessive but they only seem to take about 5 sec on my machine so it's fine. Otherwise I'm also fine with all the things it changes in this PR

metadata_backend/api/operators.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
```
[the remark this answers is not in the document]
csc-felipe: My idea to leave it like this is exactly that it would be a good idea to split it up.

csc-felipe:
I used the automatic configuration generator, so it explicitly shows all the enabled checks. What I did was enable all and choose which ones are disabled. And then added the `docparams` extension so that it checks the function parameters match the docstrings.

metadata_backend/api/auth.py
```diff
@@ -66,7 +66,7 @@ async def login(self, req: Request) -> Response:
         # Redirect user to AAI
         response = web.HTTPSeeOther(session["url"])
         response.headers["Location"] = session["url"]
-        raise response
+        return response
```
blankdots, line 69: @csc-felipe i think this should have been left as a `raise`, have you tested this with an odic provider ?

this is something we cannot (or it is very difficult) mock with the integration tests, and needs to be tested with a oidc provider

csc-felipe: The current implementation works with SDS AAI, I have tried it manually, but what kind of tests did you have in mind?

The difference between `raise` and `return` is not so dramatic. afaik, it's more about how the middleware handles them than anything else.

csc-felipe: https://github.com/CSCfi/metadata-submitter/pull/559

PR #553, opened 2022-08-18
Add instructions for enabling swagger docs.

csc-felipe:
Added instructions to the README for enabling swagger docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,19 @@ $ docker run -p 5430:5430 metadata-submitter
 
 Frontend is built and added as static files to backend while building.
 
+## Swagger docs
+
+Swagger UI is already available in the production docker image. During development, you can enable it by executing `bash scripts/swagger/generate.sh`.
+
+Restart the server, and the swagger docs will be available at http://localhost:5430/swagger.
+
+**Swagger docs requirements**
+- `bash`
+- `Python 3`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -143,6 +143,12 @@ $ docker run -p 5430:5430 metadata-submitter
 
 Frontend is built and added as static files to backend while building.
 
+## Swagger docs
+
+Swagger UI is already available in the production docker image. During development, you can enable it by executing `bash scripts/swagger/generate.sh`. The script also needs Python 3 and PyYaml to be installed (already available if you installed the development dependencies above).
```
lilachic, line 148: Is it ok to run it `./scripts/swagger/generate.sh` when terminal is bash shell? Like an options for the README?

csc-felipe: Yes, it's ok. I would expect that to be implicit, and wouldn't need to explicitly write that it's possible to run executable scripts by prefixing their path with `./`.

And it doesn't really matter what shell you are running, because of the first line of the script, as long as `bash` is installed, the script should run as expected, at least in a Linux machine.

In the `Dockerfile` it was done like that. https://github.com/CSCfi/metadata-submitter/blob/689d46b84ff1971230625ac50fcc77f21b64de44/Dockerfile

lilachic (approved):
Thanks for the Swagger instructions. LGTM

PR #434, opened 2022-05-18
Refactor folder -> submission

csc-felipe:
### Description

It's confusing to have two different words to represent the same thing. This PR aims to unify the terminology regarding a `submission` in `metadata-submitter`, refactoring the word  `folder` to `submission`.

This is quite a large change, but no functionality should have changed. The biggest change is that the `/folder` endpoint changed to `/submission`.

There was already a file called `submission.py`, and it was renamed to `xml_submission.py` so that `folder.py` could be renamed to `submission.py`.

### Related issues

- Closes #411

### Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Changes Made

- `/folder` endpoint is now `/submission`
- Refactor folder -> submission

#### Renaming process
- File / python module renamed: `.submission` -> `.xml_submission`
- Python class renamed: `SubmissionAPIHandler` -> `XMLSubmissionAPIHandler`
- File / python module renamed: `.folder` -> `.submission`
- Word replaced`folder` -> `submission`
- Word replaced `SubmissionSubmission` -> `Submission`

Then manual checking which `folder`s were actually system folders. Hopefully I didn't miss any.

### Testing

- [x] Tests do not apply

PR #554, opened 2022-08-18
Fix schema endpoint returning 400

csc-felipe:
### Description

Inconsistencies in the internal name of the `submission` schema (with or without an **s** at the end), caused the `/v1/schemas/datacite` and `/v1/schemas/submission` to be broken, as it was looking for `submissions`

This change fixes the issue by ensuring that the schema (also called collection) is always `submission`.

There is no visible change to the API user. 

### Related issues

- Follow up from #434

### Type of change

### Changes Made

- Fix internal schema naming for submissions to "submission", while endpoint remains "submissions".

### Testing

- [x] Unit tests
- [x] Integration tests

### Mentions

@lilachic, nice that you found this bug.


csc-jm (approved):
I guess the tests just fail because of unrelated issues?

lilachic (approved):
LGTM,  Cypress e2e tests passing.

PR #556, opened 2022-08-22
Map submission `subjects` to Metax `field of science`

### LINKED ISSUE #506: Map Metadata-submitter `fields of science` to Metax `subjects`
#### Description

Metax metadata is missing mapping of `subject` to metax `field_of_science`. 
Submitter uses  OECD (The Organisation for Economic Co-operation and Development) classification for field of science. This can be found from e.g. https://www.oecd-ilibrary.org/docserver/9789264239012-en.pdf?expires=1656661002&id=id&accname=guest&checksum=1898ACACEF918D92EB0B19A17FF61AFD page 59.
Metax uses research codes from https://koodistot.suomi.fi/codescheme;registryCode=research;schemeCode=Tieteenala2010. Unfortunatly Metax do not validate other classification than YSO.
After some research there seemed to be not very good possibilities for mapping of those two.

Datacite has no validation for dedicated classification but sugests to use controlled vocalabury. SD Submit could start using same classification as Metax which is based on [Simple Knowledge Organization System](https://www.w3.org/2004/02/skos/) (SKOS) -model.
(Sources: 
https://support.datacite.org/docs/datacite-metadata-schema-44
https://schema.datacite.org/meta/kernel-4.4/doc/DataCite-MetadataKernel_v4.4.pdf
https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#subject
https://www.kiwi.fi/display/Finto/YSOn+tietomalli)

Proposition would be to change submitters classification to use elasticsearch endpoint provided by Metax https://metax.fairdata.fi/es/reference_data/field_of_science/_search?pretty=true for choosing field of science for submission on frontend as prefetched value. For API users there could be instructions to check values from http://finto.fi/okm-tieteenala/en/index.
The metadata for field_of_science is constructed as followed:
```
"field_of_science": [
    {
        "identifier": "http://www.yso.fi/onto/okm-tieteenala/ta4111",
        "pref_label": {"en": "Agronomy},
    }
]
```
#### DoD (Definition of Done)

Published dataset discoverable in Etsin will have information on field of science.

#### Testing

Integration tests.

csc-jm:
### Description

PR adds functionality to map `subjects` from submission doi info to Metax `field_of_science`. As discussed in #327 and further described in #506, the idea is to revert from using the OECD classification for subject terms and instead use the YSO fields of science. Hence there are some changes to `submission.json`.

### Related issues

Fixes #506 

### Type of change

- [x] New feature

### Changes Made

- Added a new metax mapping method
- Added new values to the metax reference fetching script
- Made the metax reference file to be a static file in this repository
- Updated the `submission.json` schema and `specification.yml` of docs
- Methods in `submission.py` add more subject info to `doiInfo`
- Added some tests
- Fixed some spelling errors

### Testing

- [x] Unit Tests
- [x] Integration Tests


scripts/metax_mappings/create_metax_references.py
```diff
@@ -69,5 +69,15 @@
         pass
 log.error(f"{n_matches} languages matched, and {no_matches} didn't")
 
+
+# parsing codes and identifiers for mapping fields of science
+fos_codes_url = "https://metax.demo.fairdata.fi/es/reference_data/field_of_science/_search?size=100"
```
csc-felipe, line 74: Are you getting all of them when `size=100`?

csc-felipe:
Please check with @hannyle if it affects the frontend. And if necessary, create a ticket there.

scripts/metax_mappings/create_metax_references.py
csc-jm: There are currently 74 different FOS in the metax reference data list so that query `size` gets them all. Not sure if they'll ever add more

metadata_backend/helpers/schemas/submission.json
```diff
@@ -148,67 +148,114 @@
...
+                                    "611 - Philosophy",
+                                    "6121 - Languages",
+                                    "6122 - Literature studies",
+                                    "6131 - Theatre, dance, music, other performing arts",
+                                    "6132 - Visual arts and design",
+                                    "614 - Theology",
+                                    "615 - History and archaeology",
+                                    "616 - Other humanities",
+                                    "9 - Other",
+                                    "999 - Other"
```
csc-jm, line 234: If the frontend utilizes this list of values for creating submissions, then this shouldn't break anything

csc-felipe:
Please, add an entry to the changelog.

csc-felipe (changes requested):
Looks great!

I'd still like to see an entry in the changelog, and it would be better if the generated file would be included in the repo as part of a separate PR.

csc-jm:
Changelog updated and the generated file was removed from this PR. However the unit tests I wrote for this PR fail now, but I swear they work if you generate the references :D 

(hence this could be merged and the unit tests will pass soon, once we add the generated file back)

csc-jm:
Ok so I cleaned up this PR with the magic of interactive rebasing. I also commented out the unit tests that are included here because they don't work without the static reference files which are added in #568. PR #568 is based on this PR so the unit tests written here pass in that PR.

csc-felipe (approved):

csc-felipe:
Nice! It's a bit funny how you made the tests in this PR, but run them only later, but I can live with it :)

metadata_backend/services/metax_service_handler.py
```diff
@@ -224,7 +225,13 @@ async def update_dataset_with_doi_info(self, datacite_info: Dict, _metax_ids: Li
 
             # Map fields from doi info to Metax schema
             mapper = MetaDataMapper(metax_id["schema"], metax_data["research_dataset"], datacite_info)
-            mapped_metax_data = mapper.map_metadata()
+            try:
+                mapped_metax_data = mapper.map_metadata()
+            except SubjectNotFoundException as error:
```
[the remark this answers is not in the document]
csc-jm: Yea I originally didn't have this because I too thought that the subjects should always correspond with the schema. But then I figured you can technically still feed the method faulty values which would cause issues. Will it ever happen? Maybe not, but if it ever does, this avoids returning a 500 response.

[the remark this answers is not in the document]
csc-felipe: Yes, makes sense.

blankdots (changes requested):
this fixes metax but I think it breaks a bit integration with datacite as the FOS is something recommended by datacite along with https://id.loc.gov/authorities/subjects.html

This requires changes to docs `specification.yml` but i marked it as a request for change until we change the change viable for datacite as well

metadata_backend/helpers/schemas/submission.json
blankdots: it was hardcoded as FOS https://github.com/CSCfi/metadata-submitter-frontend/blob/master/src/features/wizardSubmissionFolderSlice.tsx#L227-L228

csc-felipe:
> this fixes metax but I think it breaks a bit integration with datacite as the FOS is something recommended by datacite along with https://id.loc.gov/authorities/subjects.html

Could you elaborate on the link? I didn't see how is it connected to the subjects we are using.

According to Datacite specification on the metadata schema, the `subject` field is a free text field, with the possibility to indicate the scheme and URI of the resource we used. It seems pretty clear that we don't need to be limited to using FOS or LCSH. I am no specialist in research and what are valid subject classifications, but it does seem to me that we do have an alternative, and perhaps more detail could be added to the PR description about it.
image
Page 18-19 of https://schema.datacite.org/meta/kernel-4/doc/DataCite-MetadataKernel_v4.4.pdf

csc-felipe:
Or then what we should do, is manually map from Datacite to Metax. It's not so many, so it's quite doable.

metadata_backend/helpers/schemas/submission.json
blankdots: we can fix that in a later iteration of front-end, and probably we can alleviate that by making some changes in the backend

blankdots (changes requested):
i managed to come to terms with using something else than FOS, after checking it is ok, however still some changes required:

i think the `test_doi.json`, `subjects` should contain only:
```
  "subjects": [
    {
      "subject": "999 - Other",
    }
  ]
  ```
  
  and the rest:
```
      "subjectScheme": "Korkeakoulujen tutkimustiedonkeruussa käytettävä tieteenalaluokitus",
      "schemeUri": "http://www.yso.fi/onto/okm-tieteenala/conceptscheme",
      "valueUri": "http://www.yso.fi/onto/okm-tieteenala/ta999",
      "classificationCode": "999"
```

is added from the code in submissions.py as part of `_prepare_datacite_study` and `_prepare_datacite_dataset`

File: metadata_backend/helpers/schemas/submission.json (hunk 2 above)
Comment (blankdots) [reply metadata_backend/helpers/schemas/submission.json:234]: it is good at is, we can address changes when front-end starts to work on it

File: metadata_backend/helpers/schemas/submission.json [hunk 4]
```diff
@@ -148,67 +148,114 @@
                     "description": "Subject, classification code, or key phrase describing the resources specified by OECD Fields of Science and Technology (FOS)",
```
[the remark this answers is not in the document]
csc-jm: Changed

csc-jm:
Updated the `specification.yml` with the new subject names/codes, fixed the descriptions in the `submission.json` schema, moved the subject info to be assigned in `_prepare_datacite_study` and `_prepare_datacite_dataset` and rebased with `develop` branch.

blankdots (approved):

PR #568, opened 2022-09-01
Add third party resources as static files 

### LINKED ISSUE #555: Remove dependency on lexvo.org and other 3rd party resources
#### Describe the bug

Building and running tests depend on `lexvo.org`, but it's been offline, and then nothing works any more. 

#### Tasks
1. We should instead include third party files in the repo
2. Have the scripts available that can update them
3. GH actions could run the scripts automatically, check for changes, and create a PR for us.

#### Definition of Done
At least tasks 1 and 2.

csc-jm:
### Description

It was determined that the metax references should exist in the repository (We currently fetch resources from 3 different places for metax mapping and one of them went offline recently for a few days which kind of broke our tests in this repo). So this PR adds them as 3 separate static files and the script for fetching them. Also this PR enables the unit tests for metax mapper written in #556 to work now.

### Related issues

Fixes #555 
(at least tasks 1 and 2)

### Type of change

### Changes Made

- update the metax reference creation methods
- include the resulting reference files in the conf
- update `metax_mapper` methods affected by this
- activate the metax mapper unit tests that were commented out

### Testing

- [x] Unit tests


metadata_backend/conf/metax_references/fields_of_science.json
```diff
@@ -0,0 +1,670 @@
+{
+    "fields_of_science": {
```
csc-felipe, line 2: I think we could remove this, and make the whole object as fields of science.

metadata_backend/conf/metax_references/identifier_types.json
```diff
@@ -0,0 +1,25 @@
+{
+    "identifier_types": {
```
csc-felipe, line 2: I think we could remove this, and make the whole object as identifier types.

metadata_backend/conf/metax_references/languages.json
```diff
@@ -0,0 +1,7824 @@
+{
+    "languages": {
```
csc-felipe, line 2: I think we could remove this, and make the whole object as languages.

metadata_backend/services/metax_mapper.py
```diff
@@ -235,11 +235,15 @@ def __init__(self, object_type: str, metax_data: Dict, data: Dict) -> None:
         self.research_dataset = metax_data
         self.datacite_data = data
         self.affiliations: List = []
-        with open(METAX_REFERENCE_FILE, "r", encoding="utf-8") as ref_file:
-            metax_reference_data = json.load(ref_file)
-        self.identifier_types = metax_reference_data["identifier_types"]
-        self.languages = metax_reference_data["languages"]
-        self.fields_of_science = metax_reference_data["fields_of_science"]
+        for file in METAX_REFERENCE_FILES:
+            with open(file, "r", encoding="utf-8") as ref_file:
```
csc-felipe, line 239: This whole block is looking not so ideal. I don't like the check if there is a key in the JSON object. I'd rather `METAX_REFERENCE_FILES` was a dictionary for this kind of implementation.

But when thinking a bit more about it, I'd rather have the file opening and loading files into memory in `conf.py`. What do you think?

metadata_backend/conf/conf.py
```diff
@@ -183,11 +183,13 @@ def create_db_client() -> AsyncIOMotorClient:
     "catalog_pid": "urn:nbn:fi:att:data-catalog-sd",
 }
 
-METAX_REFERENCE_FILE = Path(__file__).parent.parent / "conf" / "metax_references" / "metax_references.json"
-if METAX_ENABLED and not METAX_REFERENCE_FILE.is_file():
-    raise RuntimeError(
-        "You must generate the metax references to run submitter: `bash scripts/metax_mappings/fetch_refs.sh`"
-    )
+file_names = ["identifier_types.json", "languages.json", "fields_of_science.json"]
+METAX_REFERENCE_FILES = [Path(__file__).parent.parent / "conf" / "metax_references" / file for file in file_names]
```
csc-felipe, line 187: While this is nice, I think we could have a more verbose version of this command here, for readability's sake. For e.g., with loading the files here. Should probably keep the error checking. Please, notice I wrote incomplete code.
```python
METAX_REFERENCE_ROOT = Path(__file__).parent.parent / "conf" / "metax_references"
# ...
for ref_file in file_names:
    with open(METAX_REFERENCE_ROOT / ref_file) as file:
        METAX_REFERENCE[ref_file.replace(".json", "")] = file.read()
```

csc-felipe (changes requested):
I think we could move the opening of files to `conf.py`, and flatten one level of the JSON files.

Otherwise, looks good to me.

metadata_backend/services/metax_mapper.py
csc-jm: Good point

metadata_backend/conf/conf.py
csc-jm: Thanks for the example

csc-felipe (approved):

blankdots (approved):
this is fine, however the changes in #556 seem to need more attention for datacite, as this metadata emphasizes now metax integration which not all submissions will need, and if metax is require i remember datacite being the go to point for creating DOIs and that has a different way of handling the `subjects`, see #556 for more info

PR #569, opened 2022-09-02
Parse age_at_extraction value to integer

### LINKED ISSUE #557: XML parsers should set age_at_extraction value to integer
#### Describe the bug
The parser returns sample age_at_extraction value as string

#### Expected behavior
Return value as int instead of string

#### Additional context
This is needed to simplify queries in bp beacon.

csc-jm:
### Description

This PR changes the `age_at_extraction` value found in BP samples (added in #491) from string to integer.

### Related issues

Fixes #557 

### Type of change

### Changes Made

- Added a check for the `age_at_extraction` value and simple test to prove it works

### Testing

- [x] Unit Tests

### DIFF
--- a/metadata_backend/helpers/schemas/bp_bpsample.json
+++ b/metadata_backend/helpers/schemas/bp_bpsample.json
@@ -55,7 +55,7 @@
                     "title": "Tag title"
                 },
                 "value": {
-                    "type": "string",
+                    "type": ["string", "number"],
                     "title": "Description"
                 },
                 "units": {
--- a/metadata_backend/helpers/parser.py
+++ b/metadata_backend/helpers/parser.py
@@ -408,6 +408,12 @@ class XMLToJSONParser:
         slides = data["slide"] if "slide" in data else []
         slides = _separate_samples(slides, "slide")
 
+        # Turn age_at_extraction value to int
+        for specimen in specimens:
+            for attribute in specimen["specimen"]["attributes"]["attribute"]:
+                if attribute["tag"] == "age_at_extraction":
+                    attribute["value"] = int(attribute["value"])
+
         # Return all samples as an array under bpsample schema name
         samples: List[Dict] = bio_beings + specimens + blocks + slides
         return {"bpsample": samples}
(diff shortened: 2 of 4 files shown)


metadata_backend/helpers/schemas/bp_bpsample.json
```diff
@@ -55,7 +55,7 @@
                     "title": "Tag title"
                 },
                 "value": {
-                    "type": "string",
+                    "type": ["string", "integer"],
```
blankdots, line 58: having values like this might be difficult for front-end to render forms from json, that is if we are rendering forms from it.

if we do this for one specific attribute then we might need to handle all, could we have more information on why we need to do this type conversion?

another thing is that give that the attribute field also takes Units, consider some units might be float in value , so `number` type might be better

blankdots (changes requested):
not 100% sold we need this, i need more context

metadata_backend/helpers/schemas/bp_bpsample.json
csc-jm: I'm not 100% sure on the necessity of this either but it was easy to change. The context provided for this is in the issue this PR is related to:
```
Additional context

This is needed to simplify queries in bp beacon.
```

csc-jm: And good point about the `number` type

blankdots: > This is needed to simplify queries in bp beacon.

yeah, not sure this should be fixed on this, as it will affect the form generation in the front-end, and as I mentioned if we do this for one value we would need to do it for all attributes.

I have some workaround solution where we don't modify the schema, but instead when we insert in the database we add an optional parameter to the attribute e.g. `attributeDataType` which specifies how that attribute should be interpreted by an a service that gets data from metadata submitter, and in that way we will try to infer the data instead of having special cases for each attribute, and if we cannot infer it we default to string

blankdots: the `attributeDataType` does not need to be present in the json schema, but instead we can add it to the XML/JSON processing of the data whenever it is posted, and will be specific for `attribute` fields

blankdots: better yet instead of us inferring the data type we should ask the XML schema provider to specify it, as this alleviate us trying to determine which data type each attribute is.

csc-jm: Ok we now discovered that the XML schemas provided to us also specify numeric attributes as decimals and this `age_at_extraction` tag should go within that and not under a regular attribute (which the current test files some reason didn't do and just parsed the value into string, hence the confusion). Anyway, the parser parses the numeric attribute into a float already as is and hence this PR is now pointless.

csc-jm:
PR can be closed. This change was unnecessary

PR #570, opened 2022-09-06
Run integration tests with pytest

### LINKED ISSUE #448: Make integration tests idempotent

csc-felipe:
### Description

Using pytest gives better reports, and allow for easier debugging during development.

The [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html#fixtures) makes it easier to understand test dependencies, and to setup / teardown in a simple and controlled manner.

The tests are grouped in classes, and per fixture scope, after each class test the DB schema is recreated.

This is already a big change, so I tried to keep it small. Github's diff viewer doesn't make it so clear, but from Pycharm's diff, it's possible to see that very little changed in the tests. Look at the diff per commit.

### Related issues

- #448 

### Changes Made

- Refactor integration tests to run with pytest
- Remove `clean_db.py`, as it's not used any more
- Move unit test files under new folder `tests/unit`
- `test_bpdataset_gets_doi` was never executed before, and now it is (after some fixes)
- Split integration tests into multiple files


metadata_backend/api/handlers/submission.py
```diff
@@ -23,7 +23,7 @@ class SubmissionAPIHandler(RESTAPIIntegrationHandler):
     def _make_discovery_url(self, obj_data: Dict) -> str:
         """Make an url that points to a discovery service."""
         # TODO: Proper URL creation when metax is disabled
```
blankdots, line 25: might want to have this `TODO` as an issue not here

tests/integration/helpers.py
```diff
@@ -0,0 +1,648 @@
...
+    mock_auth_url,
+    objects_url,
+    publish_url,
+    submissions_url,
+    templates_url,
+    testfiles_root,
+    users_url,
+)
+
+logging.basicConfig(format=FORMAT, datefmt=DATE_FORMAT)
```
blankdots, line 26: maybe we can set this once, instead of all files

tests/integration/run_tests.py
```diff
@@ -1,2384 +0,0 @@
...
-        async with sess.get(f"{schemas_url}/{schema}") as resp:
-            assert resp.status == expected_status, f"{resp.status} {schema}"
-
-
-async def main(url):
-    """Launch different test tasks and run them."""
-    mongo = Mongo(url)
-    database = mongo.db
-
-    async with aiohttp.ClientSession() as sess:
```
blankdots, line 2204: i think when the migration of the test happened, this test case was missed:
we have 2 users each starts their own session and each does different operations, that is why we have
```
test_user_given = "Given"
test_user_family = "Family"
test_user = "user_given@test.what"

other_test_user_given = "Mock"
other_test_user_family = "Family"
other_test_user = "mock_user@test.what"
```
 
the idea behind this test is to check that each user is separated in the db, so they don't collide and information is not assigned to a different user, and sessions are handled properly.

blankdots (commented):
first look, things that can be improved

tests/integration/run_tests.py
csc-felipe: Nice catch

metadata_backend/api/handlers/submission.py
csc-felipe: Yes, but not on this change. I created issue https://github.com/CSCfi/metadata-submitter/issues/571

tests/integration/run_tests.py
csc-felipe: Now that I took a look at it, I didn't see how it is testing those things you mentioned. Submissions are grouped by project, and not by user, so users are able to see each other's submissions anyway.

We could create an issue for testing those, but ATM, I really don't see how the change impacted the testing scope of the users.

blankdots: >  so users are able to see each other's submissions anyway.

true, however not sure they are able unless part of the same project, and we could use a test case for having separate projects and we can add that issue.

What that test did was related to sessions and it used them along with submissions to see the submissions are handled properly per session and information does go in the wrong places

blankdots: will be solved i another issue/PR

tests/integration/conf.py
```diff
@@ -0,0 +1,83 @@
...
+# to form direct contact to db with eg create_submission()
+DATABASE = os.getenv("MONGO_DATABASE", "default")
+AUTHDB = os.getenv("MONGO_AUTHDB", "admin")
+HOST = os.getenv("MONGO_HOST", "localhost:27017")
+TLS = os.getenv("MONGO_SSL", False)
+
+user_id = "current"
+test_user_given = "Given"
+test_user_family = "Family"
+test_user = "user_given@test.what"
```
blankdots, line 79: this does not appear used anywhere ?

csc-felipe: https://github.com/CSCfi/metadata-submitter/pull/570/files#diff-2afe5eeddeca5cb2c6c4f5d64f10e51f8c5903029ca3c3b0b71d2432d2d01e73R142

blankdots: had an old version locally

tests/integration/conftest.py
```diff
@@ -0,0 +1,122 @@
...
+        "name": "submission test 1",
+        "description": "submission test submission 1",
+        "projectId": project_id,
+    }
+    submission_id = await post_submission(client_logged_in, submission)
+
+    yield submission_id
+
+    try:
+        await delete_submission(client_logged_in, submission_id)
```
blankdots, line 82: maybe we can add the exception in `delete_submission` instead of here ?

I know there is a `delete_submission_publish` already and maybe we can combine it with that (in a different PR would be ok) ?

tests/integration/conf.py
```diff
@@ -0,0 +1,83 @@
...
+    ("sample", "SRS001433.json", "SRS001433.json"),
+    ("dataset", "dataset.json", "dataset.json"),
+    ("run", "ERR000076.json", "ERR000076.json"),
+    ("experiment", "ERX000119.json", "ERX000119.json"),
+    ("analysis", "ERZ266973.json", "ERZ266973.json"),
+]
+test_schemas = [
+    ("submission", 200),
+    ("study", 200),
+    # ("project", 200),
```
blankdots, line 43: not sure we ever had a use case for project, we didn't eve do a JSON schema, we might want o deprecate it :man_shrugging:

blankdots (approved):
I approve with the comments:

I would prefer updated documentation on how to work with the tests `docs/test.rst` line 38 there is a section there.

How would one go about debugging the data from the integration tests ? 
personally i used integration tests to also verify the data was added ok by looking at it in the database, now not sure how to do that data is deleted and integration tests contained scenarios that were really convenient. Can we add an option not to purge the data after a test case ?

tests/integration/conftest.py
csc-felipe: I really tried to make as few changes to the existing code as I managed. ^^

tests/integration/conf.py
csc-felipe: I don't think we would, but project is one of the DB schemas, so I do think it's nice to keep it visible.

tests/integration/conftest.py
blankdots: okay

tests/integration/conf.py
blankdots: I was talking about `ENA.project.xsd` but I assume you mean the User projects :)

blankdots:
> > Can we add an option not to purge the data after a test case ?
> 
> @blankdots, could you create a new issue for this?

done: #575

csc-jm (approved):
I have nothing else to add. Interestingly, running the entire integration test suite on my machine with this branch takes about 20 seconds longer than e.g. in the current `develop` branch (assuming all the integration tests here are from before). But the logging is so clean and also the possibility to only run a single test class is fantastic.

csc-felipe:
Updated `docs/test.rst`, and added debugging instructions for debugging tests run with `pytest`.

csc-felipe:
Added a mention to `pytest` fixtures as well.

blankdots (approved):

csc-jm (approved):

PR #573, opened 2022-09-09
Change age_at_extraction to numeric attribute in test file

csc-jm:
### Description

The age_at_extraction value should be a number and not a string and the bp common xml schema specifies a NUMERIC_ATTRIBUTE tag. The current bp sample test file wrongly assigned it as a general ATTRIBUTE so this quick change fixes that confusion.

### Related issues

Technically related to #557 

### Type of change

### Changes Made

- update bp sample test file
- update unit test
- remove some logs from integration tests

### Testing

- [x] Unit Tests


blankdots (approved):
makes sense now

teemukataja (approved):

Issue #577, opened 2022-09-13
Check for dead code 😵


csc-felipe:
#### Proposed solution

While working on #546 or #482, I noticed that some code might not be in use.

There seem to be a few tools around to check for dead code, and from my quick research, [Vulture](https://github.com/jendrikseipp/vulture) seems to be a good candidate.

While static check might give false positives, It should be investigated why the code is reported dead, and is it really not necessary, and unused?

#### Tasks
- [x] Introduce a new check for dead code
- [x] Investigate and fix each instance

#### DoD (Definition of Done)

Remove dead code from the codebase, and prevent new instances of dead code.

#### Testing

A new check should be added to `pre-commit`, `tox`, and `ci`.

blankdots:
i went over the dead code and it seems a lot of it is the cause of a faulty rebase or rollback e.g. https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/api/handlers/user.py#L21

PR #579, opened 2022-09-13
fix Any typing and cleanup code

### LINKED ISSUE #389: Fix typing Any
#### Description
Flake8-annotations has introduced a new requirement to code styling which gives an error:
`ANN401 Dynamically typed expressions (typing.Any) are disallowed `
This is now fixed by ignoring this error in tox.ini.

It would be desirable to fix this.

#### DoD (Definition of Done)
Flake8 check is not giving ANN401 error.

#### Testing
running tox command without `ignore: ANN401`

### LINKED ISSUE #575: Bring back option to run integration tests without clean-up
#### Proposed solution

with #448 integration tests delete data without leaving the option to analyze it in a successful run, to see if it is formatted correctly as a whole. 

such feature is good for :
- schema changes - mongo is a schema-less afterall
- adding new options to mongo collections
- debugging queries to mongo without much hassle to create data

#### DoD (Definition of Done)

add option to integration tests to run them without cleaning data from databse

#### Testing

integration tests can be run without cleaning data from mongo.

blankdots:
### Related issues

closes #389 
closes #575 

### Type of change

### Changes Made

1.  fix `Any` type wherever that is possible. There were some instances like ABC class where this could not be easily satisfied as well as with decorators and some custom classes we extend from `xmlschema` and `aiohttp`. However it might be that python 3.10 + might bring better options for that 
2. remove unused code, which i suspect is the result of faulty rebase or rollback of certain features + the fact that our `master` and `develop` branches are out of sync.
3. remove some `pylint` ignores that were not required e.g. i did not see `web.HTTPSuccessful` ever being raised and did not follow why it was handled in the exception middleware
4. xml object was not properly deleted from the corresponding schema in the db, and it made unnecessary calls for some objects, as the only places it was required was for deleting `objects` and `submissions`
5. add command line options for ``--nocleanup`` integ tests and docs about it
6. update pre-commit config dependencies

### Testing

- [x] Unit Tests
- [x] Integration Tests

### Mentions

this also addresses some of the dead code in #577

### DIFF
--- a/metadata_backend/services/service_handler.py
+++ b/metadata_backend/services/service_handler.py
@@ -2,7 +2,7 @@
 
 It provides a http client with optional basic auth, and requests that retry automatically and come with error handling
 """
-from typing import Any, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 from aiohttp import BasicAuth, ClientSession, ClientTimeout
 from aiohttp.web import HTTPError, HTTPGatewayTimeout, HTTPInternalServerError
@@ -24,7 +24,8 @@ class ServiceClientError(HTTPError):
     def __init__(
         self,
         status_code: int,
-        **kwargs: Any,
+        # difficult to pinpoint type
+        **kwargs: Any,  # noqa: ANN401
     ) -> None:
         """Class to raise for http client errors.
 
@@ -104,9 +105,9 @@ class ServiceHandler:
         url: URL = None,
         path: str = "",
         params: Union[str, dict] = None,
-        json_data: Any = None,
+        json_data: Union[Dict, List[Dict]] = None,
         timeout: int = 10,
-    ) -> Any:
+    ) -> Union[dict, str]:
         """Request to service REST API.
 
         :param method: HTTP method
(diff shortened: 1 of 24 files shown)


metadata_backend/services/service_handler.py
```diff
@@ -104,9 +105,9 @@ async def _request(
         url: URL = None,
         path: str = "",
         params: Union[str, dict] = None,
-        json_data: Any = None,
+        json_data: dict = None,
```
csc-felipe, line 108: I don't think this is always a `dict`.

tox.ini
```diff
@@ -8,7 +8,7 @@ asyncio_mode = auto
 [flake8]
 max-line-length = 120
 # ANN40 will be fixed with separate PR
```
csc-felipe, line 10: Could also remove this line

csc-felipe:
It's fun when integration tests fail on the `asyncio.gather`.

metadata_backend/services/service_handler.py
blankdots: well, then mypy is not perfect :)

tests/integration/conftest.py
```diff
@@ -25,6 +25,16 @@
 from tests.integration.mongo import Mongo
 
 
+def pytest_addoption(parser):
+    """Add command line options."""
+    parser.addoption(
```
csc-felipe, line 30: nice solution

csc-felipe (approved):

teemukataja (approved):

PR #580, opened 2022-09-14
add support for LSAAI groups #548

### LINKED ISSUE #548: LS AAI group support
Our group structure is based on CSC projects that come from CSC IDM. BIGPICTURE will use LS AAI instead of CSC IDM for managing group membership. Add support for managing groups(projects) from LS AAI groups also.

teemukataja:
### Description
Add support for LifeScience AAI groups for users that don't have access to CSC projects. LS groups are a non-breaking substitute for CSC projects.

### Related issues
Closes #548 

### Type of change

- [x] This change requires a documentation update ~(this needs a follow up PR)~

### Changes Made
- modified `auth.py` to be able to handle projects/groups from `userinfo` endpoint in both CSC way and LS way
- projects have a new `origin` key in database to denote where the project came from for use in workflows

### Testing

- [x] Unit Tests
- [x] Mockauth Manual Test

### Mentions

ELIXIR-CZ for helping with setting up LS groups.

### DIFF
--- a/docs/dictionary/wordlist.txt
+++ b/docs/dictionary/wordlist.txt
@@ -117,7 +117,7 @@ covid
 cp
 cpg
 crai
-createdFrom
+createdfrom
 createfromjson
 createfromxml
 createnewdraftsubmission
@@ -212,7 +212,7 @@ experimentlinks
 experimentref
 experimenttype
 externalid
-extractedFrom
+extractedfrom
 extrainfo
 fairdata
 faire
@@ -286,7 +286,7 @@ http
 httperror
 https
 httpseeother
-HTTPSuccessful
+httpsuccessful
 identifiertype
 identitypython
 ido
@@ -346,6 +346,7 @@ librarysource
 librarystrategy
 librarytype
 lifecycle
+lifescience
 limburgish
 lims
 lingala
@@ -483,6 +484,7 @@ pdb
 pdbcls
 pdf
 peerreview
+perun
 pgm
 phenome
 physicalobject
@@ -578,7 +580,7 @@ sampleattributes
 sampledata
 sampledemuxdirective
 sampledescriptor
-sampledFrom
+sampledfrom
 samplelinks
 samplename
 samplephenotype
@@ -623,7 +625,7 @@ src
 srf
 ssl
 ssrna
-StackOverflow
+stackoverflow
 stepindex
 studyabstract
 studyattribute
@@ -639,10 +641,10 @@ subjectsschema
 submissionid
 submissionslice
 submissiontype
-submitter's
 submitterdemultiplexed
 submitterid
 submitters
+submitter's
 svg
 swati
 tabix
--- a/metadata_backend/api/auth.py
+++ b/metadata_backend/api/auth.py
@@ -152,7 +152,8 @@ class AccessHandler:
             for group in session["userinfo"]["eduperson_entitlement"]:
                 projects.append(
                     {
-                        "project_name": group,
+                        # remove the oidc client information, as it's not important to the user
+                        "project_name": group.split("#")[0],
                         "origin": "lifescience",
                     }
                 )
--- a/metadata_backend/api/auth.py
+++ b/metadata_backend/api/auth.py
@@ -115,12 +115,10 @@ class AccessHandler:
             raise web.HTTPUnauthorized(reason="Invalid OIDC callback.")
 
         # User data is read from AAI /userinfo and is used to create the user model in database
-        user_data = {
+        user_data: Dict[str, Union[str, List[Dict[str, str]]]] = {
             "user_id": "",
             "real_name": f"{session['userinfo']['given_name']} {session['userinfo']['family_name']}",
-            # projects come from AAI in this form: "project1 project2 project3"
-            # if user is not affiliated to any projects the `sdSubmitProjects` key will be missing
-            "projects": session["userinfo"]["sdSubmitProjects"].split(" "),
+            "projects": [],
         }
         if "CSCUserName" in session["userinfo"]:
             user_data["user_id"] = session["userinfo"]["CSCUserName"]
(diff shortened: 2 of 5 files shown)


metadata_backend/api/auth.py
```diff
@@ -136,8 +136,27 @@ async def callback(self, req: Request) -> Response:
...
+            groups = session["userinfo"]["eduperson_entitlement"]
+            for group in groups:
+                # LSAAI groups are of format namespaces:root_group:actual_group#client
+                # and we want to get "root_group:actual_group", and any subgroups of actual_group
+                matched = re.search(f"(?<=)({self.lsaai_root}.*?)(?=#)", group)
+                if matched:
+                    projects.append(matched.group(0))
+        else:
+            # Unknown group source, keep empty
+            pass
```
teemukataja, line 156: What do we want to do if user has no projects or groups? If I recall correctly, with CSC AAI, if there are no projects, the login fails in AAI before reaching metadata-submitter, because it is configured so. But with LS AAI, do we want to let the user pass here with no groups, or raise a 400/401?

blankdots: i would raise 401, to keep it consistent with the CSC AAI, let us not introduce special cases unless we really need them.

docs/dictionary/wordlist.txt
```diff
@@ -333,6 +333,8 @@ lang
 lastmodified
 leaveaspool
 lexically
+LifeScience
+LIFESCIENCE
```
csc-jm, line 337: AFAIK, our spell check is not case sensitive so including a single `lifescience` would check for both `LifeScience` and `LIFESCIENCE`
```suggestion
lifescience
```

teemukataja: Ok, let's try that!

csc-felipe: I've been using this command to fix the word list `sort docs/dictionary/wordlist.txt | tr A-Z a-z | sort -u -o docs/dictionary/wordlist.txt`.

docs/submitter.rst
```diff
@@ -118,9 +120,10 @@ We follow the steps of the OpenID Connect protocol.
 - The OP responds with an ID Token and usually an Access Token, which are validated with configuration provided by ``OIDC_URL``.
 - The RP can send a request with the Access Token to the UserInfo Endpoint.
 - The UserInfo Endpoint returns Claims about the End-User, use claims ``sub``, ``CSCUserName`` or ``remoteUserIdentifier`` to identify the user and start a session.
+- If using LifeScience AAI as authentication provider, then ``AAI_LIFESCIENCE_ROOT_GROUP`` must be set. This configuration is not used with CSC AAI.
```
csc-felipe, line 123: Does it mean that we can only use one or the other?

metadata_backend/api/auth.py
```diff
@@ -136,8 +136,28 @@ async def callback(self, req: Request) -> Response:
...
+            for group in groups:
+                # LSAAI groups are of format namespaces:root_group:actual_group#client
+                # and we want to get "root_group:actual_group", and any subgroups of actual_group
+                matched = re.search(f"(?<=)({self.lsaai_root}.*?)(?=#)", group)
+                if matched:
+                    projects.append(matched.group(0))
+        else:
+            # No project group information received, abort, as metadata-submitter
+            # object hierarchy depends on a project group to act as owner for objects
+            raise web.HTTPUnauthorized(reason="Insufficient user profile.")
```
csc-felipe, line 157: Perhaps the message could say that the user doesn't belong to any projects that have access to SD-Submit?

teemukataja: Yeah, I was being very vague there, I'm sure we can give the actual reason

docs/submitter.rst
teemukataja: You can set this also if you're using CSC AAI, but it does nothing. The authentication source is controlled by `OIDC_URL` and there we can only have one authentication source. This new env is only used for parsing the very long strings from LS into a shorter format.

csc-felipe:
Could the commits be squashed, or merged together? There are 3 commits about changing `wordlist.txt`. `git rebase -i develop` does the trick nicely. Change `pick` to `fixup`, and reorder the commits, so it merges the one with `fixup`, with the one above that.

teemukataja:
> Could the commits be squashed, or merged together? There are 3 commits about changing `wordlist.txt`. `git rebase -i develop` does the trick nicely. Change `pick` to `fixup`, and reorder the commits, so it merges the one with `fixup`, with the one above that.

Fixup was a nice tip! :smile:

docs/submitter.rst
blankdots, line 123: what if we have the need for multiple ROOT groups ?

teemukataja: I'm sure we can change the variable to a CSV in that case. A simpler option would be to not have this config, and take the full strings, but they are quite long. Also it takes admin configuration from LS AAI side to apply a new group to a service, so there won't be an event where we suddenly get a new root group without knowing about it.

blankdots: hmm, i am wondering if we need a mapping between projects e.g. LifeScience root group and workflows developed #578 :thinking:

csc-felipe: I think I would do the workflows differently. Maybe removing the unavailable workflows from the repo completely, or another environment variable.

csc-felipe: SDS AAI also has LS support. Isn't that enough for us? So we use only SDS AAI?

csc-felipe (approved):
This is good. If the needs change, we can make a new code change for that.

teemukataja:
This current solution only allows use of either CSC projects or LS groups, but we will be getting both of them combined from SDS AAI, so I'll need to fix that. And I can do the multiple root groups support at the same time.

docs/submitter.rst
blankdots: i don't think it is decide we will use separate instances customized per workflow

csc-felipe: Then we should filter workflows based on the groups / project. It could be implemented at the workflow level, if it would be ok to make the workflow definitions including which projects to hide the workflow from. That would be a generalized way of doing it.

teemukataja: We decided to not split the group names and take them as they are, because if we supported multiple root groups, then it would be possible to hijack groups from other root groups by creating subgroups with the names of other root groups

teemukataja:
New stuff:
- no more splitting of the group name
- added origin of project/group to db, which can be used in workflows later

metadata_backend/api/auth.py
```diff
@@ -136,8 +134,36 @@ async def callback(self, req: Request) -> Response:
...
+                        "project_name": csc_project,
+                        "origin": "csc",
+                    }
+                )
+        if "eduperson_entitlement" in session["userinfo"]:
+            # LS AAI groups come in format ["group1", "group2"]
+            for group in session["userinfo"]["eduperson_entitlement"]:
+                projects.append(
+                    {
+                        "project_name": group,
```
csc-felipe, line 155: Shouldn't we remove the `#`, and what comes after, as that is not part of the name of the group?

teemukataja: I'm okay with removing that, as it's not important to the user

csc-felipe (approved):

blankdots (approved):
approved as is.

when we have better understanding on how to operate with both we will make new issues

PR #582, opened 2022-09-19
Validate schemas

### LINKED ISSUE #581: Lint JSON schemas
#### Proposed solution

Schemas are a central part on which metadata submitter is built on, however, so far we don't check that the JSON schemas conform to the specification, or to which specification.

In #578, a new `pre-commit` check was added, and we can use its `check-metaschema` to validate our schemas.

#### Tasks
- [x] Use `check-metaschema` pre-commit hook

```
- repo: https://github.com/python-jsonschema/check-jsonschema
  rev: 0.18.2
  hooks:
    - id: check-metaschema
      files: ^metadata_backend/helpers/schemas/.*\.json$
```
- [x] Fix any discovered issues
- [x] Add a `$schema` attribute to indicate the schema version of the document
- [x] Run it in Github CI to continuously validate our schemas.

#### DoD (Definition of Done)

Continuously assert that the JSON schemas conforms with the specification.

#### Testing

New tests pass

csc-felipe:
### Related issues

- Fixes #581
- Follow-up from JSON schema validator change in #579 

### Type of change

- [x] Breaking change (possibly, as the JSON schema changed) 

### Changes Made

- Update schemas to JSON Schema 2020-12
- Add validation pre-commit for JSON schemas
- Add validation step to GitHub actions (actually tox)
- Removed `$id` fields, as they [should be a resolvable `URI`](https://json-schema.org/understanding-json-schema/structuring.html#retrieval-uri), not `self`. Validation fails when they are not resolvable
- Fix BP `sampleAttribute` parsing and schema

### Testing

- [x] Unit Tests
- [x] Schema validation with `tox` and `pre-commit`, also running with GitHub workflows

### DIFF
--- a/metadata_backend/helpers/schemas/bp_bpdataset.json
+++ b/metadata_backend/helpers/schemas/bp_bpdataset.json
@@ -1,8 +1,8 @@
 {
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
     "title": "Dataset",
-    "definitions": {
+    "$defs": {
         "identifiers": {
-            "$id": "#/definitions/identifiers",
             "title": "Identifiers",
             "description": "Identifiers to be used in the International Nucleotide Sequence Database Collaboration (INSDC) namespace.",
             "type": "object",
@@ -55,7 +55,6 @@
             }
         },
         "reference": {
-            "$id": "#/definitions/reference",
             "additionalProperties": true,
             "type": "object",
             "properties": {
@@ -75,7 +74,7 @@
                     "title": "Center Namespace"
                 },
                 "identifiers": {
-                    "$ref": "#/definitions/identifiers"
+                    "$ref": "#/$defs/identifiers"
                 }
             }
         }
@@ -129,14 +128,14 @@
             "title": "Policy Reference",
             "description": "Identifies the data access policy controlling this data set.",
             "type": "object",
-            "$ref": "#/definitions/reference"
+            "$ref": "#/$defs/reference"
         },
         "imageRef": {
             "title": "Image Reference",
             "description": "Identifies the images which are part of this dataset.",
             "type": "array",
             "items": {
-                "$ref": "#/definitions/reference"
+                "$ref": "#/$defs/reference"
             }
         },
         "annotationRef": {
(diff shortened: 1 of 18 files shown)


blankdots:
small note with this, we need to make sure the forms still get rendered in the front-end as these are used as a base for the forms.

blankdots:
we might be ok: https://ajv.js.org/json-schema.html#draft-2020-12 support seems to be ok, but still need to be tested as it looks like a breaking change, which maybe i introduced when changing the validator with #579

csc-felipe:
Was it so that @blankdots will look at the schema again?

metadata_backend/helpers/schemas/bp_bpdataset.json
```diff
@@ -1,8 +1,8 @@
 {
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
```
blankdots, line 2: this line is not required

blankdots (dismissed):
small change request for now, till i get more time to go in depth with the schema and have some more insightful suggestions

metadata_backend/helpers/schemas/bp_bpdataset.json
csc-felipe: It's not required, but for future reference, it's good to know which schema version we have. It was not so easy (without prior JSON schema knowledge) to know that most of the schemas were Draft7. If they release new versions and the parsers update their default version, our tests will keep passing, because the parsers should detect the schema version from that `$schema` field.

blankdots: ok, then we can leave it.

blankdots (commented):
i looked over the schemas, i think the BP are partly ok, there are some things missing and some things that should be done differently. 
All in all i do think we need a bit of data modelling, and developing a way to reference data via aliases, as it seems the BP schemas rely heavily on that. 

I think we need a new issue for that.

blankdots:
i think the only thing that remains is to test with front-end to see if this is a breaking change or not

blankdots (approved):
i am fine with going further with this.

teemukataja (approved):

PR #583, opened 2022-09-21
Feature/token auth

### LINKED ISSUE #468: Authentication for API users
#### Description

There's a need for authentication to be implemented for API users. 
More descriptions coming soon...
   
#### DoD (Definition of Done)

A metadata submitter API user can be authenticated through API so that he can submit his metadata through an API.

#### Testing

Coming soon...

teemukataja:
### Description
Adds support for using the REST API with `Bearer` tokens issued by the configured `OIDC_URL`.

### Related issues
Closes #468 

### Type of change

- [x] This change requires a documentation update

### Changes Made

- refactored some `callback()` code in `auth.py` into functions, so that they can be re-used in the middleware
- added new case to middleware which checks for the existence of a `Bearer` token, and if one exists, it validates it at the configured `OIDC_URL` and gets back the user profile, which is used to create the user context for the duration of the request
- removed `OIDC_ENABLED` testing variable which can cause misconfiguration incidents
- set PKCE for `oidcrp` auth to be explicit
- added placeholder `DPOP` for when AAI support is ready

### Testing

- [x] Unit Tests
- [x] Integration Tests
- [x] Manual Testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Run integration tests with pytest
   - run integration tests with `--nocleanup` option 
 - Support for LifeScience groups as a substitute to CSC projects #548
+- Support for `Bearer` tokens, opening use of the API without frontend. Tokens are validated from the configured `OIDC_URL`
+- Made [PKCE](https://oidcrp.readthedocs.io/en/latest/add_on/pkce.html) settings explicit in `oidcrp` client auth
+- Added [DPOP](https://oidcrp.readthedocs.io/en/latest/add_on/dpop.html) placeholder settings for when AAI support has been implemented
 
 ### Changed
 - schema loader now matches schema files by exact match with schema #481. This means that schema file naming in metadata_backend/helpers/schemas now have rules:
@@ -52,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - remove `datacite.json` to render the form from `folder["doiInfo"]`
   - we removed `namedtype` for `contributors` and `creators` we therefore allow `additionalProperties`
   - `subjectsSchema` is a given by frontend thus we allow via `additionalProperties`
+- Removed `OIDC_ENABLED` testing variable which can cause misconfiguration incidents
 
 - remove unused code related to change from user to project ownership caused by faulty rebase or rollback of certain features #579
 
--- a/metadata_backend/api/auth.py
+++ b/metadata_backend/api/auth.py
@@ -47,6 +47,25 @@ class AccessHandler:
                     "response_types": self.auth_method.split(" "),
                     "scope": self.scope.split(" "),
                 },
+                "add_ons": {
+                    # Re-activate this once we have implemented support on AAI side
+                    # "dpop": {
+                    #     "function": "oidcrp.oauth2.add_on.dpop.add_support",
+                    #     "kwargs": {
+                    #         "signing_algorithms": [
+                    #             "ES256",
+                    #             "ES512",
+                    #         ]
+                    #     },
+                    # },
+                    "pkce": {
+                        "function": "oidcrp.oauth2.add_on.pkce.add_support",
+                        "kwargs": {
+                            "code_challenge_length": 64,
+                            "code_challenge_method": "S256",
+                        },
+                    },
+                },
             },
         }
         self.rph = RPHandler(self.oidc_url, client_configs=self.oidc_conf)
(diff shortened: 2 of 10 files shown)


metadata_backend/api/auth.py
```diff
@@ -1,7 +1,7 @@
 """Handle Access for request and OIDC workflow."""
 
 import time
-from typing import Dict, List, Union
+from typing import Any, Dict, List, Union
```
teemukataja, line 4: I know we don't like `Any`, but it comes from `oidcrp` which is oblivious to the userinfo response type `userinfo: Dict[str, Any]`, so I put this in there :smile:

csc-felipe:
There was a comment on the issue #468:

> We would need to restrict access so that users cannot create new users, or have permissions we were not expecting them to.

@teemukataja, did you have any thoughts about this?

metadata_backend/api/middlewares.py
```diff
@@ -121,3 +129,52 @@ def _json_problem(exception: web.HTTPError, url: URL, _type: str = "about:blank"
...
+    # aai = ServiceHandler(base_url=URL(oidc_config["userinfo_endpoint"]), http_client_headers=headers)
+    aai.base_url = oidc_config["userinfo_endpoint"]
+    aai.http_client_headers = headers
+    userinfo = await aai._request(method="GET")
+    if not userinfo:
+        return session  # current empty session
+
+    # Parse user profile data from userinfo endpoint response, these steps can raise 401
+    single_session: AccessHandler = AccessHandler(aai_config)
+    user_data: Dict[str, Union[List[Dict[str, str]], str]] = await single_session._create_user_data(userinfo)
```
csc-felipe, line 169: Should we have a type alias for this, as it seems to repeat? 
e.g. `UserData = Dict[str, Union[List[Dict[str, str]], str]]`

teemukataja:
> There was a comment on the issue #468:
> 
> > We would need to restrict access so that users cannot create new users, or have permissions we were not expecting them to.
> 
> @teemukataja, did you have any thoughts about this?

Bearer tokens provide the same permissions as frontend login

blankdots:
> Bearer tokens provide the same permissions as frontend login

I would prefer if we would think of a way to restrict them, only to known users that either exist in the db or are added (e.g. by an admin in the db)

teemukataja:
> > Bearer tokens provide the same permissions as frontend login
> 
> I would prefer if we would think of a way to restrict them, only to known users that either exist in the db or are added (e.g. by an admin in the db)

So a user can use the API if they have either first logged in via frontend, or had an admin manually add their username to the db?

blankdots:
> So I'm thinking why does the API need to be more restricted than the frontend?

at the moment it does not.

The only scenarios i could think of is to prevent abuse, e.g. automatic deletion and submission, but that can be done in front-end as well.
We can leave it like this, and we can figure out particular use cases on top of it, if it needs restricting.

csc-felipe (approved):

CHANGELOG.md
```diff
@@ -25,6 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Run integration tests with pytest
   - run integration tests with `--nocleanup` option 
 - Support for LifeScience groups as a substitute to CSC projects #548
+- Support for `Bearer` tokens, opening use of the API without frontend. Tokens are validated from the configured `OIDC_URL`
+- Made [PKCE](https://oidcrp.readthedocs.io/en/latest/add_on/pkce.html) settings explicit in `oidcrp` client auth
+- Added [DPOP](https://oidcrp.readthedocs.io/en/latest/add_on/dpop.html) placeholder settings for when AAI support has been implemented
+- Removed `OIDC_ENABLED` testing variable which can cause misconfiguration incidents
```
blankdots: this should go under removed section

metadata_backend/api/auth.py
```diff
@@ -264,3 +243,69 @@ async def _set_user(
...
+        """Parse user profile data from userinfo endpoint response.
+
+        :param userinfo: dict from userinfo containing user profile
+        :returns: parsed user profile data containing username and real name
+        """
+        # User data is read from AAI /userinfo and is used to create the user model in database
+        user_data: UserData = {
+            "user_id": "",
+            "real_name": f"{userinfo['given_name']} {userinfo['family_name']}",
+            "projects": [],
```
blankdots, line 258: should this be a `ProjectList` type as used in other places as well ?

blankdots (changes requested):

metadata_backend/api/auth.py
blankdots: my bad this was defined already and did not see it.

blankdots (approved):

PR #584, opened 2022-09-21
Submission and its objects not updateable after publishing

### LINKED ISSUE #516: Make adding objects to submission after publishing not possible
#### Description

At the moment it is possible to add new objects to a submission after it has been published. The object is added to submission metadata objects but it cannot be published again. This causes strange behavior in SD Submit and conflicts with its principles.

#### DoD (Definition of Done)

It is not possible for API users to add a new object to a submission after it has been published.
 
#### Testing

Integration and unit tests.

csc-jm:
### Description

This PR ensures that submissions and the objects it contains cannot be modified or deleted after it has entered `published` state, nor can new objects be added to submission's objects. Also some `pre-commit` generated changes included, which can be removed if they seem pointless.

### Related issues

Fixes #516 

### Type of change

### Changes Made

- added checks to post/patch endpoints to make sure published submission is not edited
- added some tests to ensure the issue is solved
- added pre-commit generated changes to some files

### Testing

- [x] Unit Tests
- [x] Integration Tests


tests/integration/test_submissions.py
```diff
@@ -322,6 +322,25 @@ async def test_crud_submissions_works(self, client_logged_in, project_id):
             assert res["drafts"] == [], "there are drafts in submission, expected empty"
             assert len(res["metadataObjects"]) == 4, "submission metadataObjects content mismatch"
 
+        # check that submission cannot be updated and objects cannot be added into published submission
```
csc-felipe, line 325: Objects should not be able to be updated or deleted, either. Is it already tested?

blankdots: - delete should be checked here - https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/api/handlers/object.py#L246
- updated here: https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/api/handlers/object.py#L332

though not sure we have integration or unit tests for these

blankdots (commented):
in operator delete_submission we already check https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/api/operators.py#L934

but we also check it here: https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/api/handlers/submission.py#L714

maybe we don't need to do the check twice ?

metadata_backend/api/handlers/object.py
```diff
@@ -120,6 +120,15 @@ async def post_object(self, req: Request) -> Response:
...
 
+        # check if submission is already published
+        # objects shouldn't be added to published submissions
+        try:
+            await submission_op.check_submission_published(submission_id)
+        except web.HTTPException as exc:
+            reason = f"Submission with id {submission_id} is published and more objects cannot be added to it."
+            LOG.error(reason)
+            raise web.HTTPUnauthorized(reason=reason) from exc
+
```
blankdots, line 131: i think a simple `await submission_op.check_submission_published(submission_id)` would do as the `check_submission_published` raises `HTTPUnauthorized`

csc-jm: Yea, it just gives a different reason with the `HTTPException`. But I was also thinking of just altering the `check_submission_published` reason string to something more general

csc-felipe:
@csc-jm was there any reason that you made this as a Draft PR?

csc-jm:
Yea I was still in the middle of testing other things for this PR, hence still draft PR. I also ran `pre-commit run --all-fil
es` on my end and included all the resulted changes in this. Not sure if those are needed tho

tests/integration/test_submissions.py
csc-jm: These should all be tested now.

csc-jm:
Probably should have tagged this as `WIP` when I was still drafting it, but the previous comments should resolved now.

blankdots (commented):
there are some
```
if exists:
            if published:
```
without the `else` branch for the first if, maybe we can have `if exists and published` ?

csc-felipe:
> there are some
> 
> ```
> if exists:
>             if published:
> ```

I didn't see these, where were they?

csc-felipe:
> Probably should have tagged this as `WIP` when I was still drafting it, but the previous comments should resolved now.

Either that, or a mention somewhere could be good. I was wondering before if you made it as draft PR without intending to do so, as it wasn't clear for me.

csc-felipe:
I see that the DoD of the issue is only about preventing adding objects to published submissions, but I wonder if it should be the scope that any changes to a submission are not allowed, or do we make a new issue for that?

I can still PUT doi and dac. Also, trying to PUT a study doesn't necessarily return `401` (but can't replace the study either). And trying to publish a published submission also doesn't give `401`.

blankdots:
> I didn't see these, where were they?

- https://github.com/CSCfi/metadata-submitter/blob/bugfix/dont-add-objects-after-publishing/metadata_backend/api/handlers/object.py#L337-L338
- https://github.com/CSCfi/metadata-submitter/blob/bugfix/dont-add-objects-after-publishing/metadata_backend/api/handlers/object.py#L394-L395
- https://github.com/CSCfi/metadata-submitter/blob/bugfix/dont-add-objects-after-publishing/metadata_backend/api/handlers/xml_submission.py#L235-L236
- this https://github.com/CSCfi/metadata-submitter/blob/bugfix/dont-add-objects-after-publishing/metadata_backend/api/handlers/restapi.py#L94-L95 probably not required any more ?

these lines https://github.com/CSCfi/metadata-submitter/blob/bugfix/dont-add-objects-after-publishing/metadata_backend/api/handlers/object.py#L251-L263  appear to have an else branch, for the `if exists`

csc-jm:
Those `if exists:` and `if published:` statements are combined now and added more published checks so that objects can't be replaced either and re-publishing doesn't cause unwanted failures either as pointed out by @csc-felipe

metadata_backend/api/handlers/restapi.py
```diff
@@ -91,8 +91,6 @@ async def _handle_check_ownership(self, req: Request, collection: str, accession
 
             submission_op = SubmissionOperator(db_client)
             check, submission_id, _ = await submission_op.check_object_in_submission(collection, accession_id)
-            # if published:
-            #     _check = True
```
csc-jm, line 95: I couldn't figure out a reason for having this. Do we need to have it when it was already commented out?

blankdots:
> object now results in 401 but I don't know if 400 would be more appropriate.

401 is most used when a client provides invalid credentials. 403 could be used as the client does not have enough privileges, credentials or another reason to do the current action, Also 401 header requires a `WWW-Authenticate` header.

see recap section at: https://auth0.com/blog/amp/forbidden-unauthorized-http-status-codes/

I would go with 403 ( i would recommend this as it tells that this is not allowed) or 400.

metadata_backend/api/handlers/restapi.py
blankdots: i vote to remove

csc-felipe:
I think that the `403` makes more sense, as it clearly indicates the action is not allowed. While `400` would indicate that the user made a mistake (which is partly true).

Makes me wonder if `405` would be more fitting here.

[405 Method Not Allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405)

> The request method is known by the server but is not supported by the target resource. For example, an API may not allow calling DELETE to remove a resource.

blankdots:
> Makes me wonder if `405` would be more fitting here. thinking

`405` is method not allowed which is HTTP method, but we allow that method but we don't allow to do operation on that specific object so i would personally vote for `403`, as not sure what other thing would be ok there (i was looking at `409` but I haven't seen it anywhere besides a `PUT` or `417` but not sure what to write in `Expect` header)
with `405` we have to specify in a header `Allow` methods

csc-felipe:
Indeed, the `405` needs an extra header with the allowed methods, and according to the documentation, those are the currently allowed methods for the target resource. So it still sounds like it would be a fitting response ^^. Albeit, one that needs extra work to implement.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405

blankdots:
> Albeit, one that needs extra work to implement.

ok then i am ok with 405, but we need that `Allow` header, and I think we don't allow anything to published resources.

csc-felipe:
> we don't allow anything to published resources.

At least `GET` should be allowed, no?

blankdots:
> > we don't allow anything to published resources.
> 
> At least `GET` should be allowed, no?
yeah, that is the base, I meant any methods that imply changes. my bad/

csc-jm:
I suggest we close this PR since the DoD is technically already met and open a new issue to switch the `401` responses to `405`

blankdots (approved):

csc-felipe (approved):

PR #585, opened 2022-09-22
add healthcheck for connecting services

### LINKED ISSUE #572: Check connection to services with the health endpoint
#### Proposed solution

We have `check_connection` method in `service_handler`. The response from that check for each service we currently have could be nice to include in the `/health` endpoint.

#### DoD (Definition of Done)

Service connection statuses are included in the `/health` endpoint.

#### Testing

Unit test (?)

blankdots:
### Description

- datacite has a heartbeat endpoint that responds with `200` or `500`
- rems has a health endpoint with details about the status of the installation
- metax does not currently have any health endpoint so we do a HEAD request to base URL

### Related issues

Close #572

### Type of change

### Changes Made

* make the `DOI_API` URL default to base DOI URL instead of the ending with `/dois` 
* add datacite, rems and metax health checks in the API response also based on performance indicators (API timed reponses)
* modify mock datacite, rems and metax endpoints to answer for healthchecks

### Testing

- [x] Unit Tests
- [x] Integration Tests

### DIFF
--- a/metadata_backend/api/health.py
+++ b/metadata_backend/api/health.py
@@ -8,12 +8,30 @@ from aiohttp.web import Request, Response
 from motor.motor_asyncio import AsyncIOMotorClient
 from pymongo.errors import ConnectionFailure
 
+from metadata_backend.api.auth import AAIServiceHandler
+
 from ..conf.conf import url
 from ..helpers.logger import LOG
+from ..services.datacite_service_handler import DataciteServiceHandler
+from ..services.metax_service_handler import MetaxServiceHandler
+from ..services.rems_service_handler import RemsServiceHandler
 
 
 class HealthHandler:
     """Handler for health check."""
+
+    def __init__(
+        self,
+        metax_handler: MetaxServiceHandler,
+        datacite_handler: DataciteServiceHandler,
+        rems_handler: RemsServiceHandler,
+        aai_handler: AAIServiceHandler,
+    ) -> None:
+        """Endpoints should have access to metax and datacite services."""
+        self.metax_handler = metax_handler
+        self.datacite_handler = datacite_handler
+        self.rems_handler = rems_handler
+        self.aai_handler = aai_handler
 
     async def get_health_status(self, _: Request) -> Response:
         """Check health status of the application and return a JSON object portraying the status.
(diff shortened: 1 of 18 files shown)


teemukataja (approved):

blankdots:
so there might be a health endpoint for metax: https://github.com/CSCfi/fairdata-metax/issues/6 - putting this back as draft till we know what responses we get

metadata_backend/api/health.py
```diff
@@ -30,9 +33,31 @@ async def get_health_status(self, _: Request) -> Response:
             services["database"] = {"status": "Ok"} if conn < 1000 else {"status": "Degraded"}
         else:
             services["database"] = {"status": "Down"}
+
+        # Determine the status of loaded services
+
+        metax_handler = MetaxServiceHandler()
+        datacite_handler = DataciteServiceHandler()
+        rems_handler = RemsServiceHandler()
```
csc-felipe, line 41: I have tried to keep only one instance of the service handlers, passing them to the relevant classes in `__init__`. Perhaps it would be nice to keep with the consistency?

metadata_backend/services/datacite_service_handler.py
```diff
@@ -128,5 +129,36 @@ async def delete(self, doi: str) -> None:
...
+                    status = "Down"
+
+                return {"status": status}
+        except Exception as e:
+
+            LOG.info(f"Datacite REST API status retrieval failed with {e}.")
+            return {"status": "Error"}
+
+        finally:
+            await self._client.close()
```
csc-felipe, line 164: In the approach described above, we'd have one client session per service handler, and the client should be closed in the server teardown process, instead of after each request.

csc-felipe (changes requested):
Now there are two new and differently name methods in the service handlers, `heartbeat` and `healthcheck`. Could they have the same name, and could an `abstract` method be added to the `ServiceHandler` class, as the idea seems to be that most or all of them have health checks?

Also, would it work if the exception catching would look for `web.HTTPError`, instead of the broad `Exception`?

blankdots:
> Could they have the same name, and could an `abstract` method be added to the `ServiceHandler` class, as the idea seems to be that most or all of them have health checks?

not really as they have different ways of checking for the health and some are just heart beats or pinging the server. For datacite i need to check the response and status, for Rems i need to check the healthy response, for metax we need to ping it.

csc-felipe (approved):

csc-felipe:
Oh no, I messed up! How come the tests are failing now?

blankdots:
> Oh no, I messed up! How come the tests are failing now?

i will revert your commit and fix stuff, I think i know what is wrong

metadata_backend/api/middlewares.py
```diff
@@ -147,9 +146,10 @@ async def create_session_with_token(req: web.Request) -> aiohttp_session.Session
     else:
         return session  # current empty session
 
-    # Get path to userinfo from OIDC config
-    aai = ServiceHandler(base_url=URL(aai_config["oidc_url"].rstrip("/")))
+    # we need to create a new instance every time, this is not something we can re-use
+    aai = AAIServiceHandler()
```
blankdots, line 150: @teemukataja if you have different opinions on how to manage this, we can change.

teemukataja: We discussed that this is probably not possible

teemukataja (approved):

PR #590, opened 2022-09-28
Automatically sort dictionary

csc-felipe:
### Description

Sorting the dictionary is a boring task, so this is an automation for that.

Be sure you `pre-commit install` to benefit from `pre-commit` hooks.

### Changes Made

- sorts alphabetically
- removes duplicates
- convert to lowercase
- removes words that are not used any more (requires `ripgrep`)

**NB:** 

The feature for removing unused words from the dictionary requires [`ripgrep`](https://github.com/BurntSushi/ripgrep). I know it's not a standard tool, but it runs faster than the alternative for this case. If it's not available, the script will skip this step.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,6 +84,14 @@ repos:
 
   - repo: local
     hooks:
+    - id: sort
+      name: sort dictionary, remove unused and duplicates
+      entry: ./scripts/sort.sh
+      language: script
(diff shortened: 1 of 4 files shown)


blankdots (approved):
no objection.

I usually use included functions in vscode to sort and format a selection.

csc-jm (approved):
Seems handy

Issue #588, opened 2022-09-26
Unify DAC and policy implementations


csc-felipe:
#### Proposed solution

Currently we have two distinct implementations for DAC and policy.
1. From EGA schema
2. From REMS

and they are not connected in any way.

#### Tasks
- [ ] Design unification for both implementations
The proposed solution would be to:
  1. Rename, in the submission schema `dac` to `rems` https://github.com/CSCfi/metadata-submitter/pull/648
  2. Accept EGA XML and validate against data that lives in REMS - adding EGA DAC object to DB, and configuring the `submission.rems` field
  3. When setting REMS fields through the API (including front-end), create (or replace) the EGA DAC object to DB. #532

- [ ] Make code changes
- [ ] Update tests
- [ ] Create ticket for the frontend

#### DoD (Definition of Done)

DAC and policy implementations are unified, and interchangeable.

PR #596, opened 2022-09-29
Enforce workflow definitions

### LINKED ISSUE #591: Validate submissions against their workflow
#### Proposed solution

After the introduction of `workflows` in #578, we should also check that the submission has been created according to its workflow.

#### Tasks

- [x] #596
- [ ] #588
- [x] #619 
- [x] #620 
- [x] #621

#### DoD (Definition of Done)

Submissions must follow their workflow definition, and the process is well documented.

#### Testing

Manual tests, unit, and integration tests

csc-felipe:
### Description

Initial implementation, please give some feedback, mention things I might be missing, how do you think this should be done.

#### Tasks

- [x] remove checks whether a service handler is enabled (that functionality is obsolete with workflows)
- [x] Add workflow to submission schema
- [x] Submission follows the workflow 🤓
  - [x] Step `required`
  - [x] Schema `required`
  - [x] Schema `allowMultipleObjects`
  - [x] Schema `requires`
  - [x] Schema `requires_or`
  - [x] Publish `endpoint`
  - [ ] Publish `requires` (To be implemented) - #621
  - [x] Publish `requiredSchemas`
- [x] Update tests
- [x] Write new tests
  - [x] Workflow endpoints
  - [x] Workflow parser
  - [ ] Test different submission cases, to trigger various possibilities of pass / fail (after #588 is done) - new issue #620
- [x] Write documentation (In the CHANGELOG, PR description, code, and API docs)

### Related issues

- Fixes #591
- Follow-up on #588, #620,#619

### Type of change

### Changes Made

This is quite a large change, mostly because of refactoring, but also because of changes and checks added in various places.

- Removed the possibility to disable Metax and REMS
- Submissions must have a workflow: `FEGA`, `BigPicture`, or `SDSX`
- Check that object's schema being added to a submission belong to the workflow
- New `Workflow` class for parsing and extracting data from a workflow
- Refactor `publish` endpoint into its own python `class` and `module`
- Refactored Mongo queries to return the value for a single field
- There's more clear distinction between publishing to each integration: `datacite`, `metax`, and `rems`

### Testing

- [x] Unit Tests
- [x] Integration Tests

### Mentions

@hannyle, this is the workflow breaking change PR.

Changes to DAC and Policy still need to be done to be able to fully enforce the workflows.


metadata_backend/api/handlers/object.py
```diff
@@ -153,6 +163,13 @@ async def post_object(self, req: Request) -> Response:
                 JSONValidator(content, schema_type).validate
             operator = Operator(db_client)
 
+        schemas_in_submission = set()
+        async for _, schema, _ in self.iter_submission_objects(submission):
```
blankdots, line 167: i am not 100% sure but we might be iterating twice over the objects in a submission `get_collection_objects` on line 140 in this file

metadata_backend/api/handlers/publish.py
```diff
@@ -0,0 +1,579 @@
...
+                for item in requires_or:
+                    if item in schemas_in_submission:
+                        break
+                else:
+                    reason = (
+                        f"'{schema}' object doesn't fulfill the submission's type {workflow.name} 'requires_or' "
+                        f"field. It should include at least one of '{requires_or}'"
+                    )
+                    raise web.HTTPBadRequest(reason=reason)
+        # TO_DO: Needs more work on the tests ^^
```
blankdots, line 486: it would be nice if TODO's are not in code, but as issues or check boxes in an issue

metadata_backend/api/handlers/restapi.py
```diff
@@ -220,6 +220,30 @@ def _header_links(self, url: str, page: int, size: int, total_objects: int) -> C
...
+        :param submission: Submission data
+        :param obj_op: Object Operator
+
+        yields accession_id, schema, object_data
+        """
+        for _obj in submission["metadataObjects"]:
+            accession_id = _obj["accessionId"]
+            schema = _obj["schema"]
+            if obj_op is not None:
+                object_data, _ = await obj_op.read_metadata_object(schema, accession_id)
```
blankdots, line 238: we might already have this function as `get_collection_objects` in operators, why not use that ?

I think it replaces this whole iteration ?

blankdots (dismissed):
i think these can be improved

tests/unit/test_handlers.py
```diff
@@ -1018,3 +983,58 @@ async def test_put_submission_doi_passes_and_returns_id(self):
...
+            return_value=(
+                {"id": "prefix/suffix-study", "data": {"attributes": {"url": "http://metax_id", "types": {}}}},
+                [{"id": "prefix/suffix-dataset", "data": {"attributes": {"url": "http://metax_id", "types": {}}}}],
+            ),
+        ), patch(
+            f"{self._publish_handler}.create_metax_dataset", return_value=None
+        ), self.p_get_sess_restapi:
+            response = await self.client.patch(f"{API_PREFIX}/publish/FOL12345678")
+            json_resp = await response.json()
+            print(json_resp)
```
csc-jm, line 1038: Probably not needed?

metadata_backend/helpers/schemas/workflow.json
```diff
@@ -46,7 +46,8 @@
                             "description": "The other of the schema is the order it will render in the front-end",
```
csc-jm, line 46: Noticed this sentence. Is it supposed to say `The order of the schema...`?

metadata_backend/api/handlers/object.py
blankdots: :thinking: it does happen only for `study`, but for me it seems those functions do the same thing: check how many objects of a certain schema are in a submission

metadata_backend/helpers/schemas/workflow.json
csc-felipe: It is 😅

metadata_backend/api/handlers/restapi.py
csc-felipe: I'm not sure what do you mean with _accession id filtering_ in this case. What I tried to do was an iterator to go over all objects in a submission. In some cases, all that's needed are the `schema`, in which case `accession_id` and `object_data` are not necessary, and there's no point to read from `mongo`. Then, for the cases when the `object_data` is used, it returns that as well. Could be better if it would be two different methods as well, to make it more clear, and without hidden magic tricks.

metadata_backend/api/handlers/publish.py
csc-felipe: Yes, I intend to work on it still, and make a new issue if any TODO's are left. I agree that it's good to have issues for them, and as long as we have issues for them, it could be beneficial to have `TODO` in the code to find it more easily.

metadata_backend/api/handlers/restapi.py
blankdots: I see, the maybe all that is required is  what is done at lines 234 -236 i think that is done already by `get_collection_objects`.

What i mean is that we should take advantage of mongo queries when doing this processing.

I don't view mongo queries as hidden tricks, but as means to optimize and speed up the code. There is something else to let the DB give you the formatted result to exactly what you need and we should take advantage of that not read the dict from mongo then process in python. Mongo DB has means to filter and it is optimised for that.

csc-felipe: I meant as hidden trick lines 237-238 above, that the method is doing something which is not obvious.

csc-felipe:
@blankdots you had a good perception that I did not consider the DB queries so much. I'll take a look at your suggestion, and what other queries are part of the change.

metadata_backend/api/handlers/object.py
csc-felipe: I changed, so that the iteration has explicitly no calls to DB, which should be cheaper than querying mongo.

metadata_backend/api/handlers/restapi.py
csc-felipe: While that method gets all objects in the same collection, I wanted to iterate over all `metadataObjects` of a submission. So I don't think `get_collection_objects` would fit here.

So I get the impression that explicitly getting each object by ID asynchronously is not so bad. The optimization I'd imagine here would be to use a single `mongo` query, with a cursor, that asynchronously returns each object that belongs to a submission with each iteration. What do you think?

csc-felipe:
> maybe this could be an idea, what do you think?

I think your suggestion is quite similar, but with a filtering query by schema. Instead, we'd need to filter by `submission_id`. What do you think?

The `AsyncIterator` is the same as `AsyncGenerator`, but without the `receiving` type hint. So only the yielded values are typed.

https://docs.python.org/3/library/typing.html#typing.AsyncGenerator

csc-felipe:
Perhaps calls that attempt to change a submission in a way that doesn't follow the workflow should also return `405` instead of `400`?

blankdots:
> I think your suggestion is quite similar, but with a filtering query by schema. Instead, we'd need to filter by `submission_id`. What do you think?

yeah that is why i gave it as an example because it was quite similar, and yeah i am biased towards generators but we can go with what you did.

blankdots:
> Perhaps calls that attempt to change a submission in a way that doesn't follow the workflow should also return `405` instead of `400`?

`405` is method not allowed which is HTTP method, but we allow that method but we don't allow to do operation on that specific object so i would personally go with `403`. 
with `405` we have to specify in a header `Allow` methods

csc-felipe:
> yeah that is why i gave it as an example because it was quite similar, and yeah i am biased towards generators but we can go with what you did.

I still didn't understand the comment about the generator. Would you like to elaborate?

blankdots:
> I still didn't understand the comment about the generator. Would you like to elaborate?

i like using generators, personal choice that is all

csc-felipe:
> i like using generators, personal choice that is all

Didn't you like the one I made? 😢

blankdots:
> Didn't you like the one I made? cry

i said i do, so no need to change, all is good :smile:

metadata_backend/api/handlers/restapi.py
blankdots: sounds ok.

metadata_backend/api/handlers/publish.py
csc-felipe: This needs not only tests, but also the DAC and Policy changes from #588.

metadata_backend/helpers/workflow.py
```diff
@@ -0,0 +1,103 @@
+"""Utilities for processing a workflow."""
```
blankdots, line 1: i am starting to think based on what we do here, we might as well put the workflows in a collection, if need to check these many things on them.

For the schemas we did not need that much processing, that is why we don't have them in a mongodb collection, however i think for the workflows it might make sense. and yes we can use json schemas to validate them in mongodb https://www.mongodb.com/docs/manual/core/schema-validation/specify-json-schema/#std-label-schema-validation-json

This will eliminate the need of this custom class and processing we do on workflow JSONs in python.

For now just a thought.

csc-felipe: I don't see how it would reduce the need for processing.

One idea I had later was to create a single instance of each workflow.

The processing should happen only once, as the methods are marked as property, they should be cached by Python.

blankdots: to make it makes more sense to ask using a db query if objects of a certain schema are allowed multiple objects or not e.g.
we get a schema -> query.allowed_objects_workflow(schema)-> true/false

blankdots: but this is not the subject of this PR, was just an idea.

csc-felipe: Pushed some changes. What do you think?

metadata_backend/api/operators.py
```diff
@@ -707,32 +707,83 @@ def __init__(self, db_client: AsyncIOMotorClient) -> None:
...
+
+        reason = (
+            f"Submission: '{submission_id}' has an invalid {field}, "
+            f"expected 'str', got {type(value)}. This is a bug."
+        )
+        LOG.error(reason)
+        raise web.HTTPInternalServerError(reason=reason)
+
+    async def get_submission_field_list(self, submission_id: str, field: str) -> list:
+        """Get the only a dict field from the submission.
```
csc-jm, line 763: Same docstring confusion here

metadata_backend/api/operators.py
```diff
@@ -707,32 +707,83 @@ def __init__(self, db_client: AsyncIOMotorClient) -> None:
...
+        value = await self.get_submission_field(submission_id, field)
+        if isinstance(value, str):
+            return value
+
+        reason = (
+            f"Submission: '{submission_id}' has an invalid {field}, "
+            f"expected 'str', got {type(value)}. This is a bug."
+        )
+        LOG.error(reason)
+        raise web.HTTPInternalServerError(reason=reason)
```
csc-jm, line 760: Based on what happens in `get_submission_field()`, I'm not sure this edge case can even be reached but might as well have it just in case.

tests/unit/test_workflows.py
```diff
@@ -49,5 +57,139 @@ def test_referenced_schemas_exist(self):
...
+                                "required": True,
+                                "allowMultipleObjects": True,
+                            },
+                        ],
+                    }
+                ],
+                "publish": [{"name": "datacite", "endpoint": "datacite", "requiredSchemas": ["datacite"]}],
+            }
+        ),
+    }
```
csc-jm, line 150: Perhaps these into a test file instead? It can be here too, I don't mind that much

csc-jm (commented):
Few things I noticed. Nothing too crucial tho

csc-felipe:
Rebased

metadata_backend/api/operators.py
csc-felipe: That error should be raised if you'd try to get a different type of field with that method. So the method expected a string, but you asked for a boolean. The exception will show there was a mistake.

tests/unit/test_workflows.py
csc-felipe: Done

metadata_backend/api/operators.py
csc-felipe: Fixed

metadata_backend/api/operators.py
```diff
@@ -707,32 +707,83 @@ def __init__(self, db_client: AsyncIOMotorClient) -> None:
...
-            LOG.error(reason)
-            raise web.HTTPBadRequest(reason=reason)
+                raise web.HTTPBadRequest(reason=reason) from error
+
+        reason = f"Submission: '{submission_id}' not found."
+        LOG.error(reason)
+        raise web.HTTPBadRequest(reason=reason)
+
+    async def get_submission_field_str(self, submission_id: str, field: str) -> str:
+        """Get the only a string field from the submission.
```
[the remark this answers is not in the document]
csc-felipe: Fixed

metadata_backend/api/operators.py
```diff
@@ -707,32 +707,83 @@ def __init__(self, db_client: AsyncIOMotorClient) -> None:
         """
         self.db_service = DBService(mongo_database, db_client)
 
-    async def get_submission_project(self, submission_id: str) -> str:
-        """Get the project ID the submission is associated to.
+    async def get_submission_field(self, submission_id: str, field: str) -> Union[str, list, dict]:
+        """Get the only a field from the submission.
```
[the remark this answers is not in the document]
csc-felipe: Fixed

csc-felipe:
Fixed @csc-jm comments

blankdots (approved):
i went again over the code, for now i don't have any other suggestions. I think it very good as it is.

Note: i thought i had some suggestions, however after understanding how things were done, i think it should be as it is.

PR #597, opened 2022-09-29
Refactor/logging fmt

### LINKED ISSUE #576: Format log strings as recommended best practice
#### Proposed solution

Log strings with interpolated variables are only interpolated when the log statement is triggered.
Using f-strings in log statements mean that they are always interpolated, regardless of log levels.
`Pylint` has a check for f-strings in log statements to be replaced with, for e.g. `log.debug("That value goes here %s", value_to_interpolate)`.

We can also maintain the readability by setting the formatter to use `{}` instead of the default `%`. [As seen in this SO](https://stackoverflow.com/questions/56269302/how-to-setup-python-logging-format-using-f-string-style#58714589).

References
- [Pylint `logging-fstring-interpolation`](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html)
- [Python logging](https://docs.python.org/3/library/logging.html?highlight=logging#logging.Logger.debug)
- [Logging formatter](https://docs.python.org/3/library/logging.html#logging.Formatter)

#### DoD (Definition of Done)

- Enable Pylint `logging-fstring-interpolation` checks, and fix all occurrences.
- Change default logging formatter to `{`

#### Testing

Pylint checks pass

blankdots:
### Related issues

closes #576 

### Type of change

### Changes Made

- standardize logging under Exception to `log.exception` instead of `log.error` or `log.info`
- reformat some of the log messages to be clearer 
- enable pylint checking for `logging-fstring-interpolation`
- switch logging style to `{` 
  - Note: not sure this helped with anything but was part of the issue DoD

- bonus pylint checks:
  - `fixme` enabled - there are issues to address those TODO
  - `useless-param-doc` enabled, some changes were required for rems and metax service handlers
  - `supress-message` enabled

### Testing

- [x] Tests do not apply


csc-felipe (changes requested):
This is nice!

I noticed a couple of typos, and was wondering if `'%s'` could become `%r`.

blankdots:
> I noticed a couple of typos, and was wondering if `'%s'` could become `%r`.

where ?

csc-felipe:
> where ?, only that one place

Everywhere there's a quoted `%s`?
`%r` is quoting the values, so I think it's nicer than quoting `%s` as `'%s'`.

csc-felipe:
I was under the illusion that using `style="{"` would allow to format the logging messages more similar to `f-strings`. Sorry 😅

csc-felipe (approved):
👍

teemukataja (approved):

PR #618, opened 2022-10-11
Change published submission responses to 405

### LINKED ISSUE #617: Switch `401` responses to `405` when attempting to modify published submissions
#### Description

After the fixes made in #584, it was determined that the [405 Method Not Allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405) is a more appropriate response than the current [401 Unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) when POST/PATCH/PUT/DELETE requests are done to published submissions. Most of these endpoints have [this check](https://github.com/CSCfi/metadata-submitter/blob/bugfix/dont-add-objects-after-publishing/metadata_backend/api/operators.py#L971-980) so changing the `web.HTTPUnauthorized` to `web.HTTPMethodNotAllowed` mainly solves this issue.

#### DoD (Definition of Done)

Publish a submission and ensure any changes, updates, additions and deletions cause a `405` and not `401`

#### Testing

Unit and integration tests (the ones affected by this change)

csc-jm:
### Description

Since published submissions and its objects cannot be modfied or deleted, the post/patch/put/delete methods raised 401. It was determined that 405 is more suitable HTTP response for  these situations so that's what this PR changes.

### Related issues

Fixes #617 

### Type of change

### Changes Made

- changed `web.HTTPUnathorized` to `web.HTTPMethodNotAllowed` in few places
- updated specification docs
- fixed some docstrings
- updated related unit and integration tests

### Testing

- [x] Unit Tests
- [x] Integration Tests

### DIFF
--- a/metadata_backend/api/handlers/object.py
+++ b/metadata_backend/api/handlers/object.py
@@ -124,7 +124,7 @@ class ObjectAPIHandler(RESTAPIIntegrationHandler):
 
         # check if submission is already published
         # objects shouldn't be added to published submissions
-        await submission_op.check_submission_published(submission_id)
+        await submission_op.check_submission_published(submission_id, req.method)
 
         # we need to check if there is already a study in a submission
         # we only allow one study per submission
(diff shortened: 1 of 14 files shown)


metadata_backend/api/handlers/object.py
```diff
@@ -253,19 +253,14 @@ async def delete_object(self, req: Request) -> web.HTTPNoContent:
...
-            await submission_op.remove_object(submission_id, collection, accession_id)
-            _now = int(datetime.now().timestamp())
-            lastModified = {"op": "replace", "path": "/lastModified", "value": _now}
-            await submission_op.update_submission(submission_id, [lastModified])
-        else:
-            reason = "This object does not seem to belong to any user."
+        if exists and published:
+            reason = "Published objects cannot be deleted."
             LOG.error(reason)
-            raise web.HTTPUnprocessableEntity(reason=reason)
```
csc-jm, line 268: I didn't see a reason why the `else` portion could ever be reached when there is a `check_exists()` method call already right before this. There aren't even any tests that would reach this edge case currently. Hence the deletion. I could be misunderstanding though...

csc-felipe: Sounds like leftovers from a previous change.

csc-felipe:
This looks good. Should we have `HEAD` in allowed methods? I checked, and the server accepts `HEAD` requests, as expected, but not `OPTIONS`.

metadata_backend/api/handlers/object.py
blankdots: the reason was objects that were not attached to any submission - It was done as a sanity check in case the db gets altered and we have accessionIDs not attached to a submission or that are attached to more than 1 submission see `check_object_in_submission`

if we remove this else branch then you might as well modify `check_object_in_submission` and remove the first element from the return tuple of that function, thus eliminating the need for the `exists` check

blankdots (changes requested):
If we deprecate that then might as well make the change: https://github.com/CSCfi/metadata-submitter/pull/618#discussion_r992999124 or create another PR and deprecate it not as part of this

metadata_backend/api/handlers/object.py
csc-jm: I made the change here as it wasn't too big of a refactor. See if it makes sense still.

csc-jm:
Added `HEAD` method to `allowed_methods` and refactored the `check_object_in_submission` method

csc-felipe (approved):
LGTM

blankdots (approved):

Issue #148, opened 2020-11-26
Integrate with SD-Connect (swift-browser-ui) to receive info about submitted files


blankdots:
#### Description

We would need to integrate with https://github.com/neicnordic/sda-pipeline  in order to be able to associate metadata to submitted files

We can use https://github.com/neicnordic/sda-orchestration/blob/master/sda_orchestrator/utils/consumer.py to consume files from inbox queue in order to associate them to a submission.

#### DoD (Definition of Done)

Integration with RabbitMQ message broker to read the messages about submitted files and their IDs as well as dataset IDs
- [x] ~integration with SD-Connect sharing API to share bucket/objects for submission~ https://github.com/CSCfi/swift-browser-ui/issues/755
- [ ] SD-Connect service for data handover to SD-Submit
   - [ ] calculate md5/sha256 for entire file (New SD-Connect issue), both encrypted and decrypted checksums
   - [ ] re-encrypt when sharing with a key sda-pipeline 
   - [ ] store information about files in SD-Submit db
   - [ ] nice to have: https://github.com/neicnordic/sensitive-data-archive/issues/378
   - [ ]  MQ client that sends files' metadata to SD-Submit queue
   - [ ] (optional) tagging of Selected Files in SD-Connect to add to the object metadata for filtering them both in SD-Connect (new Issue) and SD-Submit `X-Object-Meta-Staged-For-Publishing` (and a separate one for buckets)
- [x] integration with sda-pipeline (including the RabbitMQ) neicnordic/sda-pipeline#622 @blankdots 
- [x] new files collection is required neicnordic/sda-pipeline#623  @csc-felipe 
- [x] files operator in sd-submit backend neicnordic/sda-pipeline#627 @csc-felipe and #633 @blankdots 
- [ ] neicnordic/sda-pipeline#629
- [ ] MQ consumer
   - requires to process information from multiple vhost and multiple queues, so that we know when a file has been, verified, completed, etc.
   - look into RabbitMQ ~streams~ (not what we need) and/or fanout
   - [ ] store files information - process message related to file (upload, remove and maybe cancel)
   - [ ] publish submission and watch for file statuses 
      - [ ] process a submission and associated files successfully 
      - [ ] reflect errors ingesting files in a submission
      - [ ] figure out if cancelling should be handled in any way
- [x] integration with Allas backend to list buckets/objects that shared to a project @blankdots
    - integration with allas s3 possible
[...]

PR #622, opened 2022-10-17
MQ and sda pipeline integration

blankdots:
### Description

This handles the Message Broker Integration and `sda-pipeline` in order to consume and publish necessary messages to finalize a submission.

### Related issues

part of #148
### Type of change

### Changes Made

- add MessageBroker class with MQPublisher  and MQConsumer separate classes functionality
- add cli tool for MQConsumer so that we can deploy multiple consumers independently of the web server - we can also go rpc as illustrated here: https://github.com/eandersson/amqpstorm/blob/2.x/examples/flask_threaded_rpc_client.py however i think as a design it is better to keep the consumer(s) separate from the web server
- add message broker publishing to workflow
- add rabbitmq + default config to integration tests
- update `.pre-commit-config.yaml`
- small fixes
- refactor `file.json` to keep track of versions
- add `version` `objectid` to file in submision

### Testing

- [x] Integration Tests (at least partly)

### Mentions

Files integration will be continued in a different PR, to make it easier to review
right now the integration with sda-pipeline is minimal however full integration will be handled in a different PR


metadata_backend/helpers/schemas/workflow.json
```diff
@@ -137,7 +138,24 @@
...
+                },
+                "if": {
+                    "properties": {
+                        "name": {
+                            "const": "messageBroker"
+                        }
                     }
+                },
+                "then": {
+                    "required": [
```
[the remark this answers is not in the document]
blankdots: not sure what you see it, however this is equivalent to saying if the `name` is `messageBroker` then add another required property to the ones listed on line 102 thus making the list `["name", "endpoint", "exchange"]`

blankdots:
unit tests and full functionality for the files operator as well as sda-pipeline integration will need to be split in other PRs

csc-felipe (approved):
This is a big one. Mostly adding new code, which is not yet in use, so I expect we might need to adjust and write tests later.

On a side note, `scripts` folder starts to look funny, with lots of service configuration files. But I'm ok if we solve that some other time.

docs/dictionary/wordlist.txt
```diff
@@ -650,9 +658,9 @@ subjectsschema
 submissionid
 submissionslice
 submissiontype
+submitter's
 submitterdemultiplexed
 submitterid
-submitter's
```
csc-jm, line 655: 🤨

blankdots: well yeah smth is off in the sorting, will revert that

csc-jm: If you keep it just like that, the sorting on my end will at least never change it again :D

blankdots: i think there is a difference between how vscode sorts and how the pre-commit hook `sort.sh` sorts

csc-jm (approved):

blankdots:
merging this, as other things depend on it.

PR #625, opened 2022-10-18
Split operator classes to separate files

csc-jm:
### Description

While working on another issue where I was making changes to operators, I thought it's time to separate each class to their own file as is the case with handlers. It was also ignoring the pylint `too-many-lines` check which we have set at 1000 lines.

### Type of change

- [x] Bug fix (in the sense that the operators file doesn't have to skip the pylint check)

### Changes Made

- operator classes from `operators.py` are now under `operators/` in different files
- added an empty line to one of the test files and a way to not remove that when the integration test rebuilds that file
- updated previous unit tests to work with this change

### Testing

- [x] Unit Tests


blankdots (approved):
this make sense

note: i don't get why the `submitter's` in wordlist.txt gets switched place on your commits ? can you ignore that change so it does not get committed every time

csc-jm:
> note: i don't get why the `submitter's` in wordlist.txt gets switched place on your commits ? can you ignore that change so it does not get committed every time

Must be a pre-commit thing as well. I'm as confused as you 😅

blankdots (approved):

PR #626, opened 2022-10-18
pre-commit speed ups

csc-felipe:
- Speed up pre-commit by removing dictionary deduplication from `sort.sh`
- Speed-up unit tests by using `pytest-xdist`
- Update python deps


blankdots (commented):
changelog would be nice

PR #630, opened 2022-10-27
Add line numbers to xml validation errors

### LINKED ISSUE #91: Add line numbers to xml validation error messages
#### Proposed solution

Currently `XMLSchemaValidationError` exceptions raised by `xmlschema` library do not include specific line numbers of the error occurrence in the XML file. Including the line number in the error reason could be useful information for users and `ParseError` exceptions already include the position in the `detail.reason` that gets returned upon validation error.

@otahontas mentioned in #88 that one way to do this would be:
```
import lxml.etree as etree
root = etree.parse("XPTO.XML").getroot()
errors = [e for e in schema.iter_errors(root)]
```

#### DoD (Definition of Done)

`detail.reason` generated in xml validation method includes line number for any error case.

As an example, validating `SRP000539_invalid6.xml` could return:
`Unexpected child with tag 'BAD_ELEMENT' at position 2, line 34.`

#### Testing

Unit tests

csc-jm:
### Description

Before this PR, validation error reasons when validating an XML file would be something:
`Unexpected child with tag 'BAD_ELEMENT' at position 2.`
This PR makes those more understandable so they'd instead come out as:
`Unexpected child with tag 'BAD_ELEMENT' at line 34.`

Also the changes made here enable the validator to find all the errors from the file at once.

### Related issues

Fixes #91 

### Type of change

### Changes Made

- changed the xml validator to add line numbers to error messages
- updated existing unit tests

### Testing

- [x] Unit Tests


metadata_backend/helpers/validator.py
```diff
@@ -52,33 +56,40 @@ def resp_body(self) -> str:
...
+                # line number replaces element position which is more valuable information
+                reason = re.sub(r"position [0-9]+", f"line {line_num}", reason)
+            else:
+                # line number still added as extra info to error reason
+                reason = reason + f" (line {line_num})"
+            detail = {"reason": reason, "instance": instance}
+            response["detail"].append(detail)
+
+        response["detail"] = response["detail"][0] if len(response["detail"]) == 1 else response["detail"]
+        return response
```
csc-jm, line 91: The idea here is to include all the failing instances within the file in the same response as a list since it's possible to collect them all. Not sure if that's needed tho, but in case this is a good idea the openapi docs might need to be updated since currently the `response["detail"]` type is set as object. As I was going through the `specification.yml` I noticed there's existing unrelated error there on line 55. So I'll leave the doc update out of this for now.

csc-jm:
When researching possibilities for this, I thought I had found a simple solution with `xmlschema.assertValid()` as shown [here](https://lxml.de/validation.html#xmlschema) where you could get something like this:
```
xmlschema.assertValid(doc2)
Traceback (most recent call last):
  ...
lxml.etree.DocumentInvalid: Element 'c': This element is not expected. Expected is ( b )., line 1
```
However, after implementing this vulture complained that using lxml.etree is not safe anymore. It would have also required schemas to be loaded in a different way, which would have resulted in a bigger refactor. So the solution in this PR is best I could come up with without refactoring things too much.

docs/specification.yml
```diff
@@ -46,7 +46,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/403Forbidden"
-  /v1/submit:
+  /v1/submit/{workflow}:
```
csc-felipe, line 49: Nice find

docs/specification.yml
```diff
@@ -1785,17 +1785,20 @@ components:
       properties:
         isValid:
           type: boolean
-          description: true of false if valid. if false a detail key would be good to detail the error but not required
+          description: True or false if valid. If false, a detail key is included that details the error 
         detail:
-          type: object
-          description: details of the error if there was one.
+          description: details of the error if there were some
+          oneOf:
```
csc-felipe, line 1791: Now that I look at the code, I think `detail` is always a `str`. There was some refactoring some time ago, and IFAIK, all the return messages only set a string value for detail.

docs/specification.yml
csc-jm: All credit to @blankdots  :D

docs/specification.yml
csc-felipe: https://github.com/CSCfi/metadata-submitter/blob/55f2be308c4e74c522c1f075b7df4838f2b4f35c/metadata_backend/api/middlewares.py#L123

csc-felipe: Exception's reason is a `str` https://docs.aiohttp.org/en/stable/web_reference.html?highlight=exception#http-exceptions

metadata_backend/helpers/validator.py
csc-felipe: Given the other comment about the `detail` field, what do you think, could `detail` be a string? Likely those massages will not be parsed by a script, and even if they are, a regex could do the trick in extracting information.

docs/specification.yml
csc-felipe: You'd still need to add the specification for that field. You could check the `/v1/workflows/{workflow}` endpoint, there's a `parameters` field.

metadata_backend/helpers/validator.py
csc-felipe: Now that I realize I might be asking pretty much for a new feature, that error messages stick to the JSON Problem Details format. But what do you think, does it make sense in this case?

csc-felipe (commented):
Good work putting the line numbers there ^^ Much appreciated

The error messages are quite well standardized now (https://github.com/CSCfi/metadata-submitter/pull/433), although the validation errors were not following it. The `detail` field is always a `str`, following the JSON Problem Details specification https://www.rfc-editor.org/rfc/rfc7807. It would be nice to stick with it.

Would it make sense to change it in this PR, as you are already proposing a change to the validation error message?

csc-jm:
Ok, I altered the new functionality slightly so it collects the error messages and concatenates them in the same string of the `reason` key separated by an endline. It's not ideal but things will hopefully be fixed as the whole response will need some reformatting from its prior state. I opened the issue #641 for a follow up.

csc-felipe (approved):
lgtm

blankdots (approved):

PR #627, opened 2022-10-19
Files operator

csc-felipe:
### Description

Create a `FileOperator` to handle database operations for files, and `files` endpoint

### Related issues

- #148

### Type of change

- [x] New feature (non-breaking change which adds functionality)

### Changes Made

- Refactor operators, use new `BaseOperator` to share functionality
- Add `FileOperator`
- Add `/files` API endpoint to fetch files attached to a project
- Submission files endpoints

### Testing

- [x] Unit Tests
- [x] More tests with files - issue created #629

PR #632, opened 2022-10-31
Turn base operator class to common functions file

csc-jm:
### Description

After changes done in #627 it seems unnecessary to inherit a base class with just a single method in it for all these different operator classes so this organizes the operators similar to how handlers have a common functions file instead.

### Type of change

### Changes Made

- turned the `BaseOperator` class into a single method that most of the operators import

### Testing

- [x] Unit Tests


csc-jm:
Made this fairly quickly just because I thought it cleans up the structure slightly. If it's a bad or just an unnecessary change, feel free to close without merging

blankdots:
the change looks in line with the `handlers` module (`common.py`), and streamline things. At the same time i like the class inheritance to have one common class for the db_service.

Thinking of the future, maybe an ABC class would suit us as the `generate_accession_id` might need to be an `abstractmethod` where each object implements their own way of generating the `accessionId` - e.g. we might want to customize the uuids with prefix to specify what type of object it is, or we might want to generate for a file a uuid based on sha256 checksum, to uniquely identify each file in the system.

Issue #629, opened 2022-10-27
Proper unit and integration tests for Files


csc-felipe:
#### Description

Files endpoint were added in #627, but didn't get enough tests. We should tests that the endpoint returns correct files given pre-existing files in MongoDB.

#### Tasks
- [ ] Files in more than one project
- [ ] Listing files for a project returns only the desired files
- [ ] Add files to a submission
- [ ] Reading files in a submission gives the correct files
- [ ] Remove file from a submission
- [ ] Update file in a submission

#### DoD (Definition of Done)

Test multiple usage cases and methods.

PR #633, opened 2022-10-31
more file submission ops

blankdots:
### Related issues

related to: #148 
continuation of #627 

### Type of change

### Changes Made

- check files have the status `ready` when being read from the submission
- prevent publish if files have in submission have status `added` (added but no metadata object) or `failed` (failed in ingestion, completion, or for any other reason) - in future version we can also check if the submission actually has files
- missing docs about raise exception in files
- update status of a file
  - `/submission/<submission_id>/files` receives a list of files 
  ```
  [
      {
        "accessionId": 1,
        "version": 1,
        "status": "ready".
        "objectId": {
           "accessionId": 4,
           "schema": "study"
        }
      },
      {
        "accessionId": 2,
        "version": 1,
        "status": "ready"
      },
      {
        "accessionId": 3,
        "version": 1,
        "status": "ready"
      }
    ]
  ```
and  transforms it to mongodb query for each file in `update_file_submission` e.g.:
```
{
    "files.$.status": "ready",
    "files.$.version": 1,
    "files.$.objectId": {
      "accessionId": 4,
      "schema": "study"
    }
```

- remove file from submission when flagged as deleted, via `remove_file_submission` which is called when a file is flagged for deletion but can also be called whenever we want to remove all files with a specific accessionId or file path from one or multiple submissions
- operation to assign metadata object to file, also marking a file as ready, done via `/submission/<submission_id>/files` as one or multiple files can be marked as ready - will be used also MQ integration.
- small bug fixes and typo fixes

### Testing

### Mentions

this is dependent on #642

### DIFF
--- a/metadata_backend/api/handlers/submission.py
+++ b/metadata_backend/api/handlers/submission.py
@@ -310,7 +310,7 @@ class SubmissionAPIHandler(RESTAPIIntegrationHandler):
             JSONValidator(submission, "submission").validate
             for file in data:
                 if "accessionId" not in file:
-                    reason = f"Updating {submission_id} failed files requires accessionId."
+                    reason = f"Updating {submission_id} failed. Files require an accessionId."
                     LOG.error(reason)
                     raise web.HTTPBadRequest(reason=reason)
                 _file_accessionId = file.pop("accessionId")
@@ -396,7 +396,7 @@ class SubmissionAPIHandler(RESTAPIIntegrationHandler):
 
         Body needs to contain a list of accessionId for files.
 
-        :param req: POST request with metadata schema in the body
+        :param req: DELETE request
         :returns: HTTP No Content response
         """
         submission_id = req.match_info["submissionId"]
--- a/metadata_backend/api/handlers/submission.py
+++ b/metadata_backend/api/handlers/submission.py
@@ -390,3 +390,28 @@ class SubmissionAPIHandler(RESTAPIIntegrationHandler):
         reason = "Request does not contain a list of Objects each with `accessionId` and `version`"
         LOG.error(reason)
         raise web.HTTPBadRequest(reason=reason)
+
+    async def delete_submission_files(self, req: Request) -> Response:
+        """Remove files from a submission.
+
+        Body needs to contain a list of accessionId for files.
+
+        :param req: POST request with metadata schema in the body
+        :returns: HTTP No Content response
+        """
+        submission_id = req.match_info["submissionId"]
+        file_accession_id = req.match_info["fileId"]
+        db_client = req.app["db_client"]
+        submission_operator = SubmissionOperator(db_client)
+
+        # Check submission exists and is not already published
+        await submission_operator.check_submission_exists(submission_id)
+        await submission_operator.check_submission_published(submission_id, req.method)
+
+        await self._handle_check_ownership(req, "submission", submission_id)
+
+        file_operator = FileOperator(db_client)
+
+        await file_operator.remove_file_submission(file_accession_id, "accessionId", submission_id)
+        LOG.info("Removing file: %r from submission with ID: %r was successful.", file_accession_id, submission_id)
+        return web.HTTPNoContent()
--- a/metadata_backend/api/handlers/submission.py
+++ b/metadata_backend/api/handlers/submission.py
@@ -363,7 +363,7 @@ class SubmissionAPIHandler(RESTAPIIntegrationHandler):
         data: List[Dict] = await req.json()
 
         if all("accessionId" in d and "version" in d for d in data):
-            # add status to
+            # set status to file as added
             data = [{**item, "status": "added"} for item in data]
             await file_operator.add_files_submission(data, submission_id)
             LOG.info("Adding files to submission with ID: %r was successful.", submission_id)
--- a/metadata_backend/api/handlers/submission.py
+++ b/metadata_backend/api/handlers/submission.py
@@ -297,20 +297,37 @@ class SubmissionAPIHandler(RESTAPIIntegrationHandler):
         elif req.path.endswith("rems"):
             schema = "rems"
             await self.check_rems_ok({"rems": data})
+        elif req.path.endswith("files"):
+            schema = "files"
         else:
             raise web.HTTPNotFound(reason=f"'{req.path}' does not exist")
 
-        op = "add"
-        if schema in submission:
-            op = "replace"
-        patch = [
-            {"op": op, "path": f"/{schema}", "value": data},
-        ]
+        if schema == "files":
+            file_operator = FileOperator(db_client)
+            # we expect to get a list of dict for files
+            # that matches the json schema when added to submission object
+            submission[schema] = data
+            JSONValidator(submission, "submission").validate
+            for file in data:
+                if "accessionId" not in file:
+                    reason = f"Updating {submission_id} failed files requires accessionId."
+                    LOG.error(reason)
+                    raise web.HTTPBadRequest(reason=reason)
+                _file_accessionId = file.pop("accessionId")
+                _file_update_op = {f"files.$.{k}": v for k, v in file.items()}
+                await file_operator.update_file_submission(_file_accessionId, submission_id, _file_update_op)
+        else:
+            op = "add"
+            if schema in submission:
+                op = "replace"
+            patch = [
+                {"op": op, "path": f"/{schema}", "value": data},
+            ]
 
-        submission[schema] = data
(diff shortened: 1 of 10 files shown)


metadata_backend/api/operators/file.py
```diff
@@ -193,18 +202,61 @@ async def read_project_files(self, project_id: str) -> List[Dict]:
...
+        ]
+        try:
+            problematic_files = await self.db_service.do_aggregate("submission", aggregate_query)
+            if len(problematic_files) > 0:
+                reason = (
+                    f"There are a problematic files: {','.join([i['accessionId'] for i in problematic_files])} "
+                    f"in the submission with id: {submission_id}"
+                )
+                LOG.error(reason)
+                raise web.HTTPBadRequest(reason=reason)
```
blankdots, line 238: not sure this should be bad status, maybe 422, or smth else, any suggestions?

metadata_backend/api/operators/file.py
```diff
@@ -193,18 +202,61 @@ async def read_project_files(self, project_id: str) -> List[Dict]:
...
+                    "version": "$files.version",
+                    "status": "$files.status",
+                }
+            },
+        ]
+        try:
+            problematic_files = await self.db_service.do_aggregate("submission", aggregate_query)
+            if len(problematic_files) > 0:
+                reason = (
+                    f"There are a problematic files: {','.join([i['accessionId'] for i in problematic_files])} "
```
csc-felipe, line 234: Would the message say that the files are not in use, and should either be used or removed?

Or should we remove unlinked files in the publishing phase? ^^

metadata_backend/api/operators/file.py
csc-felipe: I think that 400 is good.



_The server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing)._

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

csc-felipe: Although the examples are quite technical, it seems to me that the other return codes are meant for more specific cases.

I'd consider "forgetting" to either remove the files or attach them to an object as a client error ^^

metadata_backend/api/operators/file.py
```diff
@@ -193,18 +202,61 @@ async def read_project_files(self, project_id: str) -> List[Dict]:
...
+        :raises: HTTPInternalServerError if db operation failed because of connection
+        or other db issue or db aggregate does not return a list
         :returns: List of files specific to a submission
         """
-        # TO_DO: add check for ready status
         aggregate_query = [
             {"$match": {"submissionId": submission_id}},
             {"$unwind": "$files"},
+            # we match only the files that have the status ready
+            {"$match": {"files.status": "ready"}},
```
csc-felipe, line 259: I'd imagine that we should be able to read submission files in any state that they might be, no?

Should it be a parameter, that if present returns only a matching subset of files, and if not present, returns all files in a submission?

metadata_backend/api/operators/file.py
blankdots: this is more of a check to log these files, however: i am thinking more also of a retry mechanism if they users keep the files.
So this function might change.

The update (remove or unlink the files) will be done via the submission API, but this needs to be an explicit action.

metadata_backend/api/operators/file.py
blankdots: ok, I can change that.

csc-felipe:
Looking good

metadata_backend/api/operators/file.py
blankdots: yes, the plan is to attach them to the error object but for that we might need #641

blankdots: fixed with #642  and https://github.com/CSCfi/metadata-submitter/pull/633/commits/90fef6fa0a2d199967eaea403949fabac1f419d6

metadata_backend/api/handlers/submission.py
```diff
@@ -300,15 +300,16 @@ async def put_submission_path(self, req: Request) -> Response:
...
-
         op = "add"
         if schema in submission:
             op = "replace"
         patch = [
             {"op": op, "path": f"/{schema}", "value": data},
         ]
+
+        submission[schema] = data
+        JSONValidator(submission, "submission").validate
```
csc-felipe, line 311: I don't really see how moving this changed anything. Did I miss something, could you explain, please?

blankdots: the `if schema in submission` checks if the `schema` is in the `submission`, and having the  `submission[schema] = data`  before the condition check means that the `schema` key will be added to the dict, thus making the op always default to `replace`.

by changing the lines we will not be adding the `schema` to the `submission` dict.

blankdots: simple example I could think of is:
```
smth = {"smth" : 1}
 
if "cool" in smth:
    print("cool")

smth["cool"] = [1, 2]
```
vs

```
smth = {"smth" : 1}
smth["cool"] = [1, 2]
if "cool" in smth:
    print("cool")
```

csc-felipe: I got it, thanks 😅

blankdots:
changed base so that mypy fixes are included

metadata_backend/database/db_service.py
```diff
@@ -280,25 +278,43 @@ async def remove(self, collection: str, accession_id: str, data_to_be_removed: U
...
+    ) -> dict:
         """Append data by to object with accessionId in collection.
 
         :param collection: Collection where document should be searched from
         :param accession_id: ID of the object/submission/user to be appended to
         :param data_to_be_addded: str or JSON representing the data that should be
         updated to removed.
         :param upsert: If the document does not exist add it
-        :returns: True if operation was successful
+        :returns: JSON after remove if operaation was successful
```
csc-jm, line 317:
```suggestion
        :returns: JSON after remove if operation was successful
```
🤓

metadata_backend/database/db_service.py
```diff
@@ -264,14 +262,14 @@ async def update_by_key_value(self, collection: str, key: str, value: str, data_
...
-    async def remove(self, collection: str, accession_id: str, data_to_be_removed: Union[str, Dict]) -> bool:
+    async def remove(self, collection: str, accession_id: str, data_to_be_removed: Union[str, Dict]) -> dict:
         """Remove element of object by its accessionId.
 
         :param collection: Collection where document should be searched from
         :param accession_id: ID of the object/submission/user to be updated
         :param data_to_be_removed: str or JSON representing the data that should be
         updated to removed.
-        :returns: True if operation was successful
+        :returns: JSON after remove if operaation was successful
```
csc-jm, line 272:
```suggestion
        :returns: JSON after remove if operation was successful
```
🤓

metadata_backend/api/handlers/submission.py
```diff
@@ -297,19 +297,37 @@ async def put_submission_path(self, req: Request) -> Response:
...
-        JSONValidator(submission, "submission").validate
+        if schema == "files":
+            file_operator = FileOperator(db_client)
+            # we expect to get a list of dict for files
+            # that matches the json schema when added to submission object
+            submission[schema] = data
+            JSONValidator(submission, "submission").validate
+            for file in data:
+                if "accessionId" not in file:
+                    reason = f"Updating {submission_id} failed files requires accessionId."
```
csc-jm, line 313:
```suggestion
                    reason = f"Updating {submission_id} failed. Files require an accessionId."
```
more coherent(?)

metadata_backend/api/handlers/submission.py
```diff
@@ -372,3 +390,28 @@ async def add_submission_files(self, req: Request) -> Response:
         reason = "Request does not contain a list of Objects each with `accessionId` and `version`"
         LOG.error(reason)
         raise web.HTTPBadRequest(reason=reason)
+
+    async def delete_submission_files(self, req: Request) -> Response:
+        """Remove files from a submission.
+
+        Body needs to contain a list of accessionId for files.
+
+        :param req: POST request with metadata schema in the body
```
csc-jm, line 399:
```suggestion
        :param req: DELETE request
```

csc-jm (commented):
Found a few things to nitpick but otherwise looks good to me

csc-jm (approved):

csc-felipe (approved):
LGTM

Need to create an issue for tests.

blankdots:
> Need to create an issue for tests.

I thought this was it: #629

Issue #641, opened 2022-11-01
Reformat Validate endpoint response


csc-jm:
#### Description

Currently validating an xml file via `/validate` returns a `200` response whether the file is valid or not. However, upon changes done in #630, it was determined that reformatting the response according to JSON Problem specification (https://www.rfc-editor.org/rfc/rfc7807) when the file is in fact invalid. This will require changes in the front-end as well.

#### DoD (Definition of Done)

The response after an invalid file gets POSTed to the `/validate` endpoint should instead look e.g. something like this:
```
{
   "type": "",
   "title": "Bad Request",
   "detail": "Faulty XML file provided."
   "validation-failures": [ {
                         "reason": "Mismatched tag on line 1.",
                         "position": "/ELEMENT_SET"
                       },
                       {
                         "reason": "something else",
                         "position": "/"}
                     ]
   "status": 400,
   "instance": "/validate"
}
```

#### Testing

Unit and integration

PR #642, opened 2022-11-03
add option to extend problem json error

blankdots:
### Description

add additional members to problem json response if the content type can be converted to dict
### Related issues

related for #641 and required by some dev in #633 

### Type of change

### Changes Made

- update pre-commit dependencies, as packages were updated this month, so that we keep in sync
- add dict members to `application/problem+json` if that exist in the HTTP exception

all that needs to be added is:
```
text=ujson.dumps({"coolContent": [{"value": 1}, {"value": 2}]}),
content_type="application/json"
```

### Testing

### DIFF
--- a/metadata_backend/api/middlewares.py
+++ b/metadata_backend/api/middlewares.py
@@ -114,16 +114,25 @@ def _json_problem(exception: web.HTTPError, url: URL, _type: str = "about:blank"
     :param _type: Url to a document describing the error
     :returns: Problem detail JSON object as a string
     """
+    _problem = {
+        # Replace type value with an URL to
+        # a custom error document when one exists
+        "type": _type,
+        "title": HTTPStatus(exception.status).phrase,
+        "detail": exception.reason,
+        "status": exception.status,
+        "instance": url.path,  # optional
+    }
+    # we require the additional members to be sent as dict
+    # so that we can easily append them to preformated response
+    if exception.text != exception.reason and exception.content_type == "application/json":
+        # we use the content to append to extend application/problem+json
+        # response, with additional members
+        # typecasting necessary for mypy
+        _problem.update(ujson.loads(str(exception.text)))
+
     body = ujson.dumps(
-        {
-            # Replace type value with an URL to
-            # a custom error document when one exists
-            "type": _type,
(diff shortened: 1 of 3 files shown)


metadata_backend/api/middlewares.py
```diff
@@ -114,16 +119,25 @@ def _json_problem(exception: web.HTTPError, url: URL, _type: str = "about:blank"
...
+        # a custom error document when one exists
+        "type": _type,
+        "title": HTTPStatus(exception.status).phrase,
+        "detail": exception.reason,
+        "status": exception.status,
+        "instance": url.path,  # optional
+    }
+    # we require the additional members to be sent as dict
+    # so that we can easily append them to preformated response
+    if exception.text != exception.reason and exception.content_type == "application/json":
```
csc-felipe, line 133: Nice hack

metadata_backend/api/middlewares.py
```diff
@@ -100,7 +100,12 @@ async def check_session(req: web.Request, handler: aiohttp_session.Handler) -> w
 
 
 def _unauthorized(reason: str) -> web.HTTPUnauthorized:
-    return web.HTTPUnauthorized(headers={"WWW-Authenticate": 'OAuth realm="/", charset="UTF-8"'}, reason=reason)
+    return web.HTTPUnauthorized(
+        headers={"WWW-Authenticate": 'OAuth realm="/", charset="UTF-8"'},
+        reason=reason,
+        text=ujson.dumps({"coolContent": [{"value": 1}, {"value": 2}]}),
```
csc-felipe, line 106: I'm guessing this shouldn't be here ^^

csc-jm (approved):

metadata_backend/api/middlewares.py
blankdots: yeah, my bad, was testing

metadata_backend/api/middlewares.py
blankdots: yeah, i recognize, could be better, but these should be enough for current needs.

csc-jm (approved):

csc-felipe (approved):

PR #648, opened 2022-11-16
Rename `submission.dac` to `submission.rems`

### LINKED ISSUE #588: Unify DAC and policy implementations

csc-jm:
### Description

Separated this task from the other tasks given in #588 for less chaotic reviewing. So this just refactors the naming scheme of the `dac` field in the `submission` object to `rems`.

### Related issues

Partially fixes #588 

### Type of change

- [x] Breaking change (probably for front-end, considering the endpoint name is altered)

### Changes Made

- change the `submission` schema
- change the PUT endpoint name to `/v1/submission/rems`
- modify existing integration tests
- edit openapi specs

### Testing

- [x] Integration Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - New rems service handler
   - New rems mock api service for integration tests
   - New API endpoint `/v1/rems` for the frontend to retrieve DAC and Policies
-  - Submission now can have a new field, `dac`, with `workflowId`, `organizationId`, and `licenses` (array of int)
-  - Published datasets have a new field `dac` with `workflowId`, `organizationId`, `resourceId`, and `catalogueItemId`
+  - Submission now can have a new field, `dac` (changed to `rems` in #648), with `workflowId`, `organizationId`, and `licenses` (array of int)
+  - Published datasets have a new field `dac` (changed to `rems` in #648) with `workflowId`, `organizationId`, `resourceId`, and `catalogueItemId`
 - Pylint static checks
 - Added mapping for languages between the submitter and Metax-service #514
 - Added mapping for subjects from submission doi info to Metax field_of_science #556
@@ -83,6 +83,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Refactor `xml_object.py` -> `object_xml.py` #627
 - Refactor operators to use a common base class called `BaseOperator` #627
 - XML validation errors now compile all the failing elements into same error message and include the line number in the error reason #630
+- Changed the submission object's `dac` field to `rems` and its subsequent endpoints similarly #648
 
 ### Removed
 
--- a/docs/specification.yml
+++ b/docs/specification.yml
@@ -1361,11 +1361,11 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/405MethodNotAllowed"
-  /v1/submissions/{submissionId}/dac:
+  /v1/submissions/{submissionId}/rems:
     put:
       tags:
         - Manage
-      summary: Update DAC for the submission with a specified submission ID.
+      summary: Update REMS (DAC) for the submission with a specified submission ID.
       parameters:
         - name: submissionId
           in: path
@@ -1377,7 +1377,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/DAC"
+              $ref: "#/components/schemas/REMS_DAC"
       responses:
         200:
           description: OK
@@ -2842,7 +2842,7 @@ components:
             type: object
           type: array
           uniqueItems: true
-    DAC:
+    REMS_DAC:
       type: object
       required:
         - workflowId
--- a/metadata_backend/api/handlers/publish.py
+++ b/metadata_backend/api/handlers/publish.py
@@ -401,9 +401,9 @@ class PublishSubmissionAPIHandler(RESTAPIIntegrationHandler):
                 }
                 rems_datasets.append(rems_ds)
 
-        org_id = submission["dac"]["organizationId"]
-        licenses = submission["dac"]["licenses"]
-        workflow_id = submission["dac"]["workflowId"]
+        org_id = submission["rems"]["organizationId"]
+        licenses = submission["rems"]["licenses"]
+        workflow_id = submission["rems"]["workflowId"]
         for ds in rems_datasets:
             resource_id = await self.rems_handler.create_resource(
                 doi=ds["doi"], organization_id=org_id, licenses=licenses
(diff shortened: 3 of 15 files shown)


CHANGELOG.md
```diff
@@ -18,8 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - New rems service handler
   - New rems mock api service for integration tests
   - New API endpoint `/v1/rems` for the frontend to retrieve DAC and Policies
-  - Submission now can have a new field, `dac`, with `workflowId`, `organizationId`, and `licenses` (array of int)
-  - Published datasets have a new field `dac` with `workflowId`, `organizationId`, `resourceId`, and `catalogueItemId`
+  - Submission now can have a new field, `dac` (changed to `rems`), with `workflowId`, `organizationId`, and `licenses` (array of int)
```
blankdots, line 21: maybe mention #648 where it was changed

docs/specification.yml
```diff
@@ -2842,7 +2842,7 @@ components:
             type: object
           type: array
           uniqueItems: true
-    DAC:
+    REMS_DAC:
```
blankdots, line 2845: this is a workflow in REMS not a dac, maybe we can keep terminology, because from REMS workflow we derive a the  DAC metadata specific to the EGA model

blankdots (commented):
looks good to me, but i think we should keep terminology consistent with the source, REMS has workflow and approvers/Handlers  make up a DAC

docs/specification.yml
csc-felipe: It is called `REMS DAC` in the submission schema. Perhaps a new issue for that?

csc-felipe (commented):
Looks good to me. Well done, finding all the occurrences to be renamed.

After Stefan's comments are resolved, please dismiss the review and request it again. ^^

docs/specification.yml
csc-jm: Mmm, I was also wondering what constitutes as "DAC" and just "workflow". As @csc-felipe mentioned, `submission.json` originally titled these as `"REMS DAC for datasets"` but perhaps this was inaccurate titling? Should all of these just be `REMS workflows` instead?

CHANGELOG.md
csc-jm: Good point

docs/specification.yml
blankdots: new issue sounds fine.

csc-jm:
While converting all these DAC namings, I also [looked at this](https://github.com/CSCfi/metadata-submitter/blob/refactor/rename-submission-dac/metadata_backend/api/handlers/publish.py#L421-L428) but wasn't sure if that needs to change or stay like that.

blankdots:
> While converting all these DAC namings, I also [looked at this](https://github.com/CSCfi/metadata-submitter/blob/refactor/rename-submission-dac/metadata_backend/api/handlers/publish.py#L421-L428) but wasn't sure if that dataset key needs to change or stay like that.

:thinking: what is the dataset key in the lines you highlighted, could we get more details ?

csc-felipe:
That's adding the REMS details to a dataset. For each dataset we create a resource, and a catalogue item in REMS. We don't use that information anywhere at the moment, but still, it should be preserved. What line 422 does is that it adds the REMS link that can be used to apply for access to the dataset to the dataset description that will be visible in Etsin.

My answer to @csc-jm would be that `dac` could be renamed to `rems` also in that case. I expect that later the frontend will display some information derived from these values.

csc-felipe:
I do think that writing **REMS DAC** makes sense, as what we're doing is translating REMS data into DAC. But I don't mind if you prefer to have different wording. The most important imo is that it's consistent, and at least somewhat accurate.

csc-felipe:
On the other hand, writing some documentation how the publishing process works would be good. I just wonder if now is a good time, or after we make the changes we've been planning.

blankdots:
> That's adding the REMS details to a `dataset`. For each `dataset` we create a `resource`, and a `catalogue item` in REMS. We don't use that information anywhere at the moment, but still, it should be preserved. What line 422 does is that it adds the REMS link that can be used to apply for access to the `dataset` to the `dataset description` that will be visible in Etsin.


hmm i see, so it is added to these object this info: https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/helpers/schemas/ena_dataset.json and https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/helpers/schemas/ena_bpdataset.json - but we also have this information in https://github.com/CSCfi/metadata-submitter/blob/refactor/rename-submission-dac/metadata_backend/helpers/schemas/submission.json#L1165 is that correct ?

My question would be what do we do with a submission that has multiple datasets each with their own worflow in REMS because a submission can have multiple datasets and for each one in rems you need to create a separate catalogue item, however https://github.com/CSCfi/metadata-submitter/blob/refactor/rename-submission-dac/metadata_backend/helpers/schemas/submission.json#L1165 allows for only one object not an array ?

blankdots (changes requested):
> On the other hand, writing some documentation how the publishing process works would be good. I just wonder if now is a good time, or after we make the changes we've been planning.

yeah, i think this needs to come with some documentation on what represents a REMS DAC and how we map concepts, otherwise this will not go well with further development

csc-felipe:
> but we also have this information in

@blankdots, not really. Most of the information should be the same, but there are values which are specific to the dataset: `resource_id` and `catalogue_item_id`. Also the URL which is build from those values.

csc-felipe:
As for the second question, in a meeting with other people, we decided that a submission has the same workflow / policy / dac for all datasets. That's why it's an object in the submission.

blankdots:
> > but we also have this information in
> 
> @blankdots, not really. Most of the information should be the same, but there are values which are specific to the dataset: `resource_id` and `catalogue_item_id`. Also the URL which is build from those values.

then we need to remove these lines:
- https://github.com/CSCfi/metadata-submitter/blob/refactor/rename-submission-dac/metadata_backend/helpers/schemas/submission.json#L1190
- https://github.com/CSCfi/metadata-submitter/blob/refactor/rename-submission-dac/metadata_backend/helpers/schemas/submission.json#L1194

csc-felipe:
Indeed, there was a bug. The issue was created :) https://github.com/CSCfi/metadata-submitter/issues/649

csc-jm:
Moved this momentarily to WIP because I need to push and test some things. Sorry if it still generated bunch unnecessary emails. I renamed the `dac` field of a dataset. I almost also fixed the issue #649 opened by @csc-felipe in this PR but wanted to clarify something first so didn't. Before this PR gets more confusing to review, we can maybe merge and follow up with more issues

blankdots (approved):
i am ok with changes so far.

Issue #645, opened 2022-11-11
Introduce BP staining schema and its conversion to json


csc-jm:
#### Proposed solution

The BP metadata schemas had some further additions recently. [Bp.staining.xsd](https://github.com/imi-bigpicture/metadata-schema/blob/main/src/BP.staining.xsd) has been added to the BP metadata schema listing and its inclusion in our backend was requested.

#### DoD (Definition of Done)

BP staining data can be processed from an xml file, converted to json and added to database.

#### Testing

Unit and integration

Issue #644, opened 2022-11-11
Introduce BP observation schema and its conversion to json


csc-jm:
#### Proposed solution

The BP metadata schemas had some further additions recently. [Bp.observation.xsd](https://github.com/imi-bigpicture/metadata-schema/blob/main/src/BP.observation.xsd) has been added to the BP metadata schema listing and its inclusion in our backend was requested.

#### DoD (Definition of Done)

BP observations data can be processed from an xml file, converted to json and added to database.

#### Testing

Unit and integration

PR #650, opened 2022-11-29
Update BP sample schema to include `CaseType`

### LINKED ISSUE #643: Add `CaseType` to BP Sample schema
#### Proposed solution

The BP metadata schemas were edited a short while ago. Apparently `CaseType` was added to the [sample schema](https://github.com/imi-bigpicture/metadata-schema/blob/main/src/BP.sample.xsd#L18-L40) (not a public repo so link might not open) and should be added to our schema version and checked that conversion works.

#### DoD (Definition of Done)

BP cases (as part of BP sample) can be processed from an xml file, converted to json and added to database.

#### Testing

Unit and integration

csc-jm:
### Description

BP sample schema had some updates so this brings the new schema and updates things in the parsing methods etc that needed to be updated. **Still missing an actual example of a `CASE` element in an XML and the ensuing json schema.**

### Related issues

Fixes #643 

### Type of change

### Testing

- [x] Tests do not apply


tests/test_files/bpsample/template_samples.xml
```diff
@@ -93,93 +93,15 @@
...
-                        <TAG>staining_compound</TAG>
-                        <VALUE>hematoxilin</VALUE>
-                    </ATTRIBUTE>
-                    <ATTRIBUTE>
-                        <TAG>staining_method</TAG>
-                        <VALUE>chemical</VALUE>
-                    </ATTRIBUTE>
-                </ATTRIBUTE_SET>
-            </ATTRIBUTE_SET>
-        </ATTRIBUTES>
```
csc-jm, line 122: These staining informations will make their return in #645

metadata_backend/helpers/schemas/BP.schema.xsd
```diff
@@ -3,9 +3,11 @@
...
     <xs:import schemaLocation="BP.dataset.xsd"/>
     <xs:import schemaLocation="BP.image.xsd"/>
+    <xs:import schemaLocation="BP.observation.xsd"/>
     <xs:import schemaLocation="BP.sample.xsd"/>
+    <xs:import schemaLocation="BP.staining.xsd"/>
     <xs:import schemaLocation="EGA.dac.xsd" namespace="SRA"/>
     <xs:import schemaLocation="EGA.policy.xsd" namespace="SRA"/>
     <xs:import schemaLocation="SRA.study.xsd" namespace="SRA"/>
-    <xs:import schemaLocation="SRA.submission.xsd" namespace="SRA"/>
+    <xs:import schemaLocation="BP.submission.xsd" namespace="SRA"/>
```
csc-felipe, line 12: Is this correct?

metadata_backend/helpers/schemas/BP.schema.xsd
```diff
@@ -3,9 +3,11 @@
     <xs:import schemaLocation="BP.annotation.xsd"/>
     <xs:import schemaLocation="BP.dataset.xsd"/>
     <xs:import schemaLocation="BP.image.xsd"/>
+    <xs:import schemaLocation="BP.observation.xsd"/>
     <xs:import schemaLocation="BP.sample.xsd"/>
+    <xs:import schemaLocation="BP.staining.xsd"/>
```
csc-felipe, line 8: I don't see either of the added imports in the repo. Is it supposed to be like this?

csc-jm: Will be added in #644 and #645

metadata_backend/helpers/schemas/BP.schema.xsd
csc-jm: Right. This `BP.schema.xsd` is just an xsd file that Bigpicture has implemented. I'm not entirely sure of its relevance in our schema directory.

csc-jm: And so they have their own submission schema but to my knowledge we don't need to implement it here.

tests/test_files/bpsample/template_samples.xml
```diff
@@ -3,12 +3,14 @@
...
             </ATTRIBUTE>
-            <ATTRIBUTE>
+            <CODED_ATTRIBUTE>
                 <TAG>animal_species</TAG>
-                <VALUE>homo_sapiens</VALUE>
-            </ATTRIBUTE>
+                <CODE>337915000</CODE>
+                <SCHEME>SNOMED-CT</SCHEME>
+                <MEANING>Homo sapiens</MEANING>
+            </CODED_ATTRIBUTE>
```
csc-jm, line 13: Might need to do some further testing with this `CODED_ATTRIBUTE` element since it looks new to me. But I can make an additional issue if it needs some adjustment. It seems to get parsed just fine for now.

csc-jm: I actually found a slight issue with this (or rather the way our json schemas are built currently to handle these attributes). I'm working to fix it in this PR.

csc-felipe:
Is this ready now? When you want a new review, It's good to dismiss stale reviews, and request them again, and resolve conversations.

csc-jm:
Yea, so I discovered a shortcoming in the BP json schemas we created previously regarding how `ATTRIBUTES` were handled. The changes made in the `bp_sample.json` fix that issue for that schema. BP image and BP dataset schemas also require the same treatment so I'll make an issue for those and fix them as well after this is reviewed and approved.

metadata_backend/helpers/schemas/BP.schema.xsd
csc-jm: But yea, I'm pretty sure we can remove this schema

csc-felipe (approved):

Issue #517, opened 2022-07-01
XML test files need to be updated


genie9:
#### Description

Submitter uses XML test files for testing. Some files are not complete with data structure and fields to be able to notice bugs in the implementation.

#### DoD (Definition of Done)

Test files have a wide range of testable elements.

#### Testing

Integration and unit tests.

blankdots:
maybe not change files, add new ones as the old ones are designed for specific use case

PR #664, opened 2022-12-12
Update BP image and dataset json schemas to be more correct

csc-jm:
### Description

It seems these JSON schemas were not implemented quite as properly as I thought as I had a second look into them. As in the schemas included some things they didn't need and didn't include some things that they should have. 

The example files we have passed the schema validation because they kind of worked with valid files but the problem I found was that invalid json files would validate against the schemas as well. Now they shouldn't.

### Related issues

Continuation to #650 

### Type of change

### Testing

- [x] Tests do not apply (albeit a lack of different test cases created this technical debt, so maybe they should apply 🤔 )


csc-felipe:
Might be nice to add some test case to this, as you mentioned

csc-jm:
Yea, in fact this technical debt was already brought up in #517 and isn't exclusive just to these BP metadata objects submitted via XML. Maybe there can be another PR that focuses solely on adding more testable cases for all the schemas at once?

csc-jm:
Changed the `image` schema name to `bpimage` as suggested by @blankdots in #665

csc-felipe (commented):
I don't like renaming the schema files, but I guess the logic for reading them needs it to be consistent, right?

Couple of nitpicks:
- There are two commits with the same name
- There are changes to renaming _image -> bpimage_ not in the commit I'd expect those changes.

csc-felipe (changes requested):
Based on the PR description, I'd also like to see those test files you mentioned included in the integration tests.

csc-jm:
- I would have preferred keeping the `image` name as is since the image xml structure has the elements defined as `<IMAGE>` whereas e.g. for bp dataset it's actually named `<BPDATASET>`. However, the bp dataset schema name is also originally `BP.dataset.xsd` and then changed to `BP.bpdataset.xsd` when brought over here because of our pre-existing schema loader logic. I think what matters is just that the content of the schemas are the same as their source.
- The two commits with similar names -> one changes the image schema and the other changes the dataset schema
- You mean `Change image naming to bpimage` commit doesn't have all the changes it should have? Or am I misunderstanding the point?
- When I was testing the functionality of these schemas again, I didn't really create new test files. I was just adding and removing different keys and values that would invalidate the file to see what passes validation and what doesn't. We now have valid test xml files for all the objects but not many test files that have invalid data or structure. I feel like if we add more test cases for this, we should add more test cases for most other schemas too and that feels like a task for another PR (e.g. #517 could cover all these)

csc-felipe:
> I would have preferred keeping the image name as is since the image xml structure has the elements defined as <IMAGE> whereas e.g. for bp dataset it's actually named <BPDATASET>. However, the bp dataset schema name is also originally BP.dataset.xsd and then changed to BP.bpdataset.xsd when brought over here because of our pre-existing schema loader logic. I think what matters is just that the content of the schemas are the same as their source.

I see, perhaps a new logic could make sense at some point, perhaps prefixing the schema name with the provider: `bp_`.

> You mean Change image naming to bpimage commit doesn't have all the changes it should have? Or am I misunderstanding the point?

Maybe I was confused, never mind. I don't see what I thought I saw any more '^^


About the extra test files, as you were working with those files, I thought it would be fine to make the tests more robust. So that we would have automated tests for what you did manually. And I don't think the other schemas will feel diminished or jealous about it. But I don't mind if they're not included.

csc-felipe (approved):

csc-jm:
Yea I would suggest fiddling with the test xml files a bit later as there's a chance BP has to change their schemas up a bit more

blankdots (approved):

PR #696, opened 2023-02-03
Parse ISO-8601 duration in bp samples

csc-jm:
### Description

It was discovered that the BP sample attribute value `age_at_extraction` is given in ISO-8601 duration format, e.g. `"P1Y1M1D"` (= 1 year, 1 month, 1 day) instead of simple integers signifying years or days or something. BP imaging beacon needs to compare these durations so it's necessary to parse these strings to integers. So in this PR that `value` is turned into an integer representing the number of days while also retaining the original value in `originalValue` key.

### Type of change

### Changes Made

- added parsing method
- updated bp sample json schema
- updated unit tests

### Testing

- [x] Unit Tests


metadata_backend/helpers/parser.py
```diff
@@ -6,6 +6,7 @@
 from typing import Any, Dict, List, Optional, Type, Union
 
 from aiohttp import web
+from isodate import isoerror, parse_duration
```
blankdots, line 9: not sure i see the need to introduce a new package, can we use datetime ?

i see `isodate` uses datetime and some alternative solutions here: https://stackoverflow.com/questions/36976138/is-there-an-easy-way-to-convert-iso-8601-duration-to-timedelta

csc-jm: I originally went looking for sophisticated scripts for the duration parse and the one in `isodate` seemed great. I can basically replicate the method behind `parse_duration()` now with some regex. I just feel it's a bit excessive to include that in our code when you could just import it from elsewhere but if we wanna rely on least amount of packages possible then I'll do it.

blankdots: i would have nothing against that if `isodate` was maintained and they did not have fixme in the code: https://github.com/gweis/isodate/blob/2de055391bfb3c46f2e6f6a1578a699f51931b31/src/isodate/isoduration.py#L66 maybe we can go with https://github.com/metomi/isodatetime

csc-jm: That `isodatetime` looks good. I'll try with that instead

csc-jm: Made the switch now to `isodatetime` from `isodate` and it works about the same. The only downside is that `isodate` would actually parse weeks as well if they were included in the duration string e.g. `P1Y1M1W1D`. `isodatetime` doesn't allow this because apparently that's not correct ISO-8601 format so whatever. Users have to learn to not mix weeks into the longer strings. At least `P1W` would work and result to 7 days in this scenario.

blankdots: i see, maybe this some feedback we can give back that the format they are giving to us should be correct ISO-8601 standard

blankdots (dismissed):

csc-felipe:
It would be nice to have a changelog entry as well

csc-jm:
Rebased with develop and updated changelog, which made the original approval stale.

blankdots (approved):

PR #789, opened 2023-08-23
Feature/mypyconfig

gimppa:
### Description
Added mypy configuration to pyproject.toml, to make programming environments among developers more consistent.

This breaks github workflows, because of increased strictness of type checking. Updating typing is work in progress.

### Type of change

- [x] New feature (breaking change which adds functionality)

### Testing

- [x] Tests do not apply

PR #790, opened 2023-08-29
Feature/update typing

gimppa:
### Description
Updated type annotations all around. Mostly changed dict and list type annotations to be Dict[something] and List[something].

### Related issues

Fixes #789 github workflows

### Type of change

### Testing

- [x] Tests do not apply

### DIFF
--- a/metadata_backend/api/handlers/rems_proxy.py
+++ b/metadata_backend/api/handlers/rems_proxy.py
@@ -1,5 +1,5 @@
 """Get and process REMS data for the frontend."""
-from typing import Any, Dict, List, TypedDict
+from typing import Any, TypedDict
 
 from aiohttp import web
 
@@ -11,8 +11,8 @@ class Organization(TypedDict):
 
     id: str
     name: str
-    workflows: List[Dict[str, str]]
-    licenses: List[Dict[str, str]]
+    workflows: list[dict[str, str]]
+    licenses: list[dict[str, str]]
 
 
 class RemsAPIHandler(RESTAPIIntegrationHandler):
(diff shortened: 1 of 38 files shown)


csc-felipe:
Why should we use `Dict` and `List` instead of `dict` and `list` in Python 3.11? Which makes me think that perhaps we [should set `python-version` for `mypy`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-python_version).

metadata_backend/api/handlers/rems_proxy.py
```diff
@@ -75,3 +78,6 @@ def add_organization(organization_id: str, organization_name: str) -> None:
             organizations[org_id]["licenses"].append({"id": lic["id"], **title})
 
         return web.json_response(data=list(organizations.values()))
+
+
+# : Dict[str, Dict[str, Dict[str, Any] | Any] | Any]
```
csc-jm, line 83: Not quite sure what this is

metadata_backend/api/handlers/restapi.py
```diff
@@ -1,7 +1,7 @@
 """Handle HTTP methods for server."""
 import json
 from math import ceil
-from typing import AsyncIterator, Dict, Iterator, List, Optional, Tuple, Union
+from typing import Any, AsyncIterator, Dict, Iterator, List, Optional, Tuple
```
csc-jm, line 4: So according to PEP 585, these would instead be
- dict
- list
- tuple
- collections.abc.AsyncIterator
- collections.abc.Iterator

However, `Any` and `Optional` would still be imported from `typing`

csc-jm (changes requested):
I think what @csc-felipe commented is a good point. If we're updating the typing, we might as well do it with the current best practice while we're at it (as long as it's not an unnecessarily long task). So while the built-in `typing` class still exists, it's been deprecated by now as proposed in PEP 585. So you would have `list`, `dict` etc. or then `collections.abc.Callable` and `collections.abc.Iterator`. 

However, it seems the types `Any`, `Optional`, `TypedDict` and `TypeVar` which we still have in our code (at least I found in this PR) are still to be imported from `typing` 🤔

metadata_backend/api/handlers/rems_proxy.py
gimppa: This is a mistake :laughing:

csc-felipe (approved):
Looks good :)

csc-felipe:
Approved, but you still have to click on the **three dots > dismiss review** from Joonatan's requested changes.

csc-jm (approved):
Now everything looks good to me as well :P

